PR #2, opened 2021-02-09
HARMONY-717: API design iterations

rwat:
Please see card [HARMONY-717](https://bugs.earthdata.nasa.gov/browse/HARMONY-717).


rwat:
Revising branch before PR merge.

rwat:
Please see card [HARMONY-717](https://bugs.earthdata.nasa.gov/browse/HARMONY-717) and  ./design/checkpoint_03/harmony_py.ipynb for the latest design iteration.

design/checkpoint_03/harmony_py.ipynb
```diff
@@ -0,0 +1,286 @@
...
+    "    temporal={'start': datetime.date(2020, 6, 1),\n",
+    "              'stop': datetime.date(2020, 6, 30)},\n",
+    "    format=Format.ZARR\n",
+    ")\n",
+    "\n",
+    "# Authentication is stored in a client object for subsequent server interaction.\n",
+    "client = Client(Authentication())\n",
+    "\n",
+    "# Validation may be performed prior to job processing; uses Harmony server-side checking.\n",
+    "client.validate(request)\n",
```
asteiker, line 60: It's easy to forget our no-op service that will return a list of data access URLs if no transformation services exist. We want to make sure a request passes validation in this case. Nothing to modify here - just wanted to note that for when this is implemented.

design/checkpoint_03/harmony_py.ipynb
```diff
@@ -0,0 +1,286 @@
...
+    "client.status(job_id)\n",
+    "\n",
+    "# Optional\n",
+    "client.cancel(job_id)\n",
+    "\n",
+    "# Retrieve results in-region; returns a generator\n",
+    "urls = client.result_urls(job_id, region=Region.US_WEST_2)\n",
+    "# - or -\n",
+    "# Download files to a local directory\n",
+    "client.download(job_id, region=Region.US_WEST_2, directory='./research', overwrite=True)\n"
```
asteiker, line 75: How does `download` work when requesting data in region? Or should this only be an option if, say `region` isn't specified in this example, implying that the user wants the https urls?

design/checkpoint_03/harmony_py.ipynb
```diff
@@ -0,0 +1,286 @@
...
+    "job_id = client.submit(request, verbose=True, async=True)\n",
+    "\n",
+    "# Optional\n",
+    "client.status(job_id)\n",
+    "\n",
+    "# Optional\n",
+    "client.cancel(job_id)\n",
+    "\n",
+    "# Retrieve results in-region; returns a generator\n",
+    "urls = client.result_urls(job_id, region=Region.US_WEST_2)\n",
```
asteiker, line 72: `region` is optional here, or would we require another kvp to specify https?

asteiker, line 72: What do you think about adding an optional `client.stac` or `client.stac_cat` here as well? Ideally Id love to see a user being able to easily grab their stac catalog from harmony-py and then use that as input directly into pystac or intake-stac.

design/checkpoint_03/harmony_py.ipynb
```diff
@@ -0,0 +1,286 @@
...
+    "[print(b) for b in brief]\n",
+    "# ['C193529903-LPDAAC_ECS', 'MOD09GQ', 'MODIS/Terra Surface Reflectance Daily L2G Global 250m SIN Grid V006']\n",
+    "# ['C193529902-LPDAAC_ECS', 'MOD09GA', 'MODIS/Terra Surface Reflectance Daily L2G Global 1km and 500m SIN Grid V006']\n",
+    "# ['C193529899-LPDAAC_ECS', 'MOD09A1', 'MODIS/Terra Surface Reflectance 8-Day L3 Global 500m SIN Grid V006']\n",
+    "# ['C193529944-LPDAAC_ECS', 'MOD09Q1', 'MODIS/Terra Surface Reflectance 8-Day L3 Global 250m SIN Grid V006']\n",
+    "# ['C193529901-LPDAAC_ECS', 'MOD09CMG', 'MODIS/Terra Surface Reflectance Daily L3 Global 0.05Deg CMG V006']\n",
+    "\n",
+    "\n",
+    "####\n",
+    "# The CMR python library does not support variable browsing at this time.\n",
```
asteiker, line 165: Nor does it support UMM-S queries, which we also want to leverage for service discovery to help inform what requests are valid.

design/checkpoint_03/harmony_py.ipynb
```diff
@@ -0,0 +1,286 @@
...
+    "####\n",
+    "\n",
+    "\n",
+    "# The output from CMR may be used as input to the Harmony python library\n",
+    "req = Request(\n",
+    "    collection=cmr_res[0],\n",
+    "    spatial=Bbox(lat=(40, 42), lon=(-107, -105)),  # could accept bbox, shapely, geojson (polygon)\n",
+    "    temporal=Temporal(start=datetime.date(2020, 6, 1), stop=datetime.date(2020, 6, 30)),\n",
+    "    format=Format.ZARR\n",
```
asteiker, line 174: A service discovery -> Harmony request example (pending cmr python lib development) would be to query services for `cmr_res[0]`:

```
import cmr.search.services as serv
cmr_serv_res = serv.search({'collection':cmr_res[0]})
format  = cmr_serv_res.SupportedOutputFormats[0]
```

`format` could then replace `Format.ZARR` in your Harmony request. That is a super simplified example, since more than one service record can exist, and format options in UMM-S are separated by input/output combination, but that's just an idea to take it one step further so that a user could grab relevant service options and feed those into the request. The same could be done for vars.

design/checkpoint_03/harmony_py.ipynb
```diff
@@ -0,0 +1,286 @@
...
+   "metadata": {}
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Sync Request\n",
+    "res = Request.submit(req, auth, sync=True)\n",
```
asteiker, line 192: Just to avoid any confusion for external review, you may want to change this and all following examples to fit with the client concept:

`res = client.submit(request, verbose=True, sync=True)`

asteiker (changes requested):
This is shaping up to be a really solid foundation. Most of my comments are just for reference for the implementation tickets, but if you could update any relevant lines from 192 on to match the client structure, that would be helpful to avoid any confusion.

design/checkpoint_03/harmony_py.ipynb
rwat: I had assumed people would want to download data no matter their location but if the library is being executed in region, should we only return a list of URLs?

design/checkpoint_03/harmony_py.ipynb
rwat: I had intended `region` to be required in order to remind a harmony-py user that location matters and the library won't try to make any guesses or use any defaults. Maybe the region options right now would be 'US_WEST_2' and 'OUT_OF_REGION' ? We can totally change that, though.

rwat: That's a great idea.

design/checkpoint_03/harmony_py.ipynb
rwat: I'd be happy to make cards to that end but I'm not familiar enough with good acceptance criteria for those as well as how long they might take. Our respective time permitting I'd like to learn more about that.

design/checkpoint_03/harmony_py.ipynb
rwat: That awesome - I'll make updates to the notebook accordingly.

design/checkpoint_03/harmony_py.ipynb
rwat: Thank you for catching that - I overlooked those after changing the summary usage.

design/checkpoint_03/harmony_py.ipynb
asteiker: Yeah this is a good point. I have been thinking about these two binary groups of users: those who are fully outside of the cloud world, and those that only want to work with data in place within the cloud. I suppose there is nothing stopping a user from being able to download data directly from the Harmony API now even if they're running in-region, so this is valid. I don't think we want to be encouraging it though (or I shuold say, we should be encouraging analysis-in-place instead), so maybe that's where Im hung up. Worth keeping this here now so we can come back to it in the development story.

design/checkpoint_03/harmony_py.ipynb
asteiker: Harmony-705 should capture services, and I think associations are implied in both Harmony-705 and Harmony-707, so I don't think there's a need to create anything additional. Just noting what else is not supported right now.

asteiker (approved):
Thanks for making those minor updates!

kbeamnsidc (approved):

PR #10, opened 2021-02-16
Harmony 716

kbeamnsidc:
Many of the changes have already been merged because in order to invoke the GitHub Actions workflows, they need to be on the default (master) branch for the repo. So some ninja merging was necessary to test. The rest of the PR includes a Makefile, fixing up some things in the workflow definition (.github/workflows/tests.yml), and updating the README. 

From what I can tell, the updated GHA workflow will be invoked properly on PRs from this repo and from forks. Those tests should run in the fork, not in this repo. There are some interesting security things (bad) that can happen if you don't do this and the tests run in this repo. So the triggering event (pull_request) is setup to run in the fork. 

The workflow runs on all the Pythons we should support (3.6 through 3.9) and saves code coverage as a build artifact. See the Actions tab on the repo for all the details.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,32 @@
+name: Tests
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
```
chris-durbin, line 10: 👍

chris-durbin (approved):

rwat (approved):

README.md
```diff
@@ -21,9 +21,9 @@ The library is available from Pypi and can be installed with pip:
 
 ## Developing
 
-Install requirements:
+Install dependencies:
 
-> pip install -r requirements/core.txt -r requirements/dev.txt
+> make install
```
bilts, line 26: Is there any ability to install a non-dev version of master using pip without picking up the dev dependencies?  Is there any utility in doing that?

bilts (approved):

README.md
kbeamnsidc: Yeah, when building the package (in order to publish to PyPI) we'll just install what's in core. I think that will be in a separate GH Action workflow specification. So for this particular workflow I think this is correct since it's running the linter and unit tests. At that time I'd either modify the install target or add a new one / some combo thereof.

kbeamnsidc: To expand on what I wrote, I think the install target installing both core and dev is fine. My hope (?) is that we can move to something like poetry which, like npm, handle dev and runtime dependencies, and can install with or without dev dependencies.

PR #12, opened 2021-02-22
Harmony 704: Authentication

rwat:
This includes the initial implementation for authentication.


Makefile
```diff
@@ -0,0 +1,20 @@
+.PHONY: install clean lint test ci docs
```
chris-durbin, line 1: I thought I reviewed a PR where Kevin already made this change?

README.md
```diff
@@ -6,6 +6,8 @@
 
 `harmony-py` is a work-in-progress, is not feature complete, and should only be used if you would like to test `harmony-py` functionality.
 
+![Python package](https://github.com/nasa/harmony-py/workflows/Python%20package/badge.svg)
```
chris-durbin, line 9: Just curious - what does this badge show?

harmony_py/auth.py
```diff
@@ -0,0 +1,197 @@
...
+            return SessionWithHeaderRedirection(cfg.EDL_USERNAME, cfg.EDL_PASSWORD)
+        else:
+            raise MissingCredentials('Authentication: No credentials found.')
+
+
+def authenticate(username: Optional[str] = None, password: Optional[str] = None,
+                 netrc_file: Optional[bool] = False,
+                 verify: Optional[bool] = True) -> Session:
+    """
+    Create are requests-futures session for authenticated HTTP calls after optionally verifying
```
chris-durbin, line 135: are typo

harmony_py/auth.py
```diff
@@ -0,0 +1,197 @@
...
+        verify (bool, optional): Whether EDL credentials will be verified.
+
+    Returns:
+        (:obj:`SessionWithHeaderRedirection`): The authenticated requests session.
+
+    :raises MissingCredentials: No credentials were specified or found.
+    :raises BadAuthentication: Incorrect credentials or unknown error.
+    """
+    edl_verification_url = cfg.EDL_VERIFICATION_URL
+    num_workers = int(cfg.NUM_REQUESTS_WORKERS)
```
chris-durbin, line 159: What is the purpose of multiple workers in the context of EDL authentication?

harmony_py/auth.py
```diff
@@ -0,0 +1,197 @@
...
+            return futures_session
+        elif result.status_code == 401:
+            raise BadAuthentication('Authentication: incorrect or missing credentials during '
+                                    'credential verification.')
+        else:
+            raise BadAuthentication('Authentication: An unknown error occurred during credential '
+                                    f'verification: HTTP {result.status_code}')
+
+
+# if __name__ == "__main__":
```
chris-durbin, line 179: I'd remove commented out code. If this is meant as an example should we put it in a Jupyter notebook or some other Python mechanism for examples?

tests/test_auth.py
```diff
@@ -0,0 +1,121 @@
+import pytest
+from requests_futures.sessions import FuturesSession
+
+from harmony_py.auth import _authenticate, _is_edl_hostname, authenticate, BadAuthentication, \
+                            MissingCredentials, SessionWithHeaderRedirection
+
+#
```
chris-durbin, line 7: Remove commented out code in this file.

examples/authentication.ipynb
```diff
@@ -0,0 +1,70 @@
...
+    "# specify both username and password\n",
+    "session = authenticate(username='myusername', password='supersecret', verify=False)\n",
+    "\n",
+    "# specify just a username, in which case prompt the user for their password\n",
+    "session = authenticate(username='myusername', verify=False)\n",
+    "\n",
+    "# or look for .netrc credentials\n",
+    "session = authenticate(netrc_file=True, verify=False)\n",
+    "\n",
+    "# Someday, this will work:\n",
```
chris-durbin, line 58: I'm not sure if we should include this, but I'm wondering if the line above should basically be this? Can we just default netrc_file to True and only let users override to say `netrc_file=False`. That way it automatically checks for a  .netrc file or env vars EDL_USERNAME and EDL_PASSWORD when you call `session = authenticate()`?

Makefile
rwat: I'm not sure what's going on there because '.PHONY' is in master / main

README.md
rwat: I'm not sure (I didn't put it there).

harmony_py/auth.py
rwat: Authentication, specifically a validated authentication, is tied to the act of communicating with a server (Harmony + EDL) via a requests session object. In order to better support concurrent harmony-py-initiated Harmony jobs we've tentatively settled on requests-futures and they require a thread pool. The NUM_REQUESTS_WORKERS relates to the number of threads created. Conflating authentication and the act of communicating with Harmony keeps things simple as we likely won't need to separate the two for other reasons.

harmony_py/auth.py
rwat: I'm sorry - that's my bad. I had meant to remove that before pushing. A python notebook was included for example purposes, too.

tests/test_auth.py
rwat: Those comments were there to convey the test organization to the reader but I can remove them.

examples/authentication.ipynb
rwat: I had wanted to make using a .netrc file a conscious decision of a harmony-py user. It just seemed like bad design to make a no-args 'authenticate()', rely on netrc. Doing it this way, a person must choose the path they wish to take: 1) supply a username and password, or 2) supply a username and be prompted for a password, or 3) explicitly state 'I want to use netrc and I know how to troubleshoot that if it doesn't work'.

We can do the automatic thing though and I can include env behavior.

chris-durbin: Other people can weigh in. I based my comment on the way the requests Python library works automatically using a .netrc without having to specify to use one.

README.md
kbeamnsidc: image

examples/authentication.ipynb
kbeamnsidc: I'm probably 66.6% in favor of defaulting to the netrc file if one exists. But I also agree that explicit is better than implicit, I don't have a strong opinion, but the default to netrc is slightly preferable with the current API. I can think of a couple tweaks to authenticate that might help. I'll comment below.

harmony_py/config.py
```diff
@@ -0,0 +1,51 @@
+import os
+
+from dotenv import load_dotenv
+from typing import cast
+
+
+load_dotenv()
```
kbeamnsidc, line 7: If this module is imported, it'll try to find a .env file and load it, which could fail and throw an exception. I'd suggest moving this out of module scope and make it explicit behavior that happens as a side-effect of creating the Config, for example.

examples/authentication.ipynb
rwat: We had a conversation yesterday resulting in this new format:
```
session = authenticate()    # Use ENV variables, then .netrc
session = authenticate(auth=(username, password))
session = authenticate(token='ABCD1234')
```
.netrc use is automatic, ENV use is automatic and happens before .netrc, auth kwarg is used for (username, password) tuple of str, and token is a kwarg to a BearerToken str.

kbeamnsidc (dismissed):

kbeamnsidc (approved):
I pushed up a couple commits to fix unit tests, and I resolved a merge conflict in the Makefile. I'll merge this into my branch and remove my temp Session I was using for authn.

kbeamnsidc:
I forgot again: when I run `make docs` I don't see any Python docs...was that supposed to just work?

examples/authentication.ipynb
```diff
@@ -0,0 +1,70 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# note: to run, you may want to create a soft link to the library for now: `ln -s ../harmony_py/ harmony_py`\n",
+    "# note: the keyword argument `verify` defaults to True, meaning it will verify the credentials are valid during session instantiation.\n",
+    "\n",
+    "from harmony_py.auth import authenticate\n",
+    "\n",
+    "# specify both username and password\n",
+    "session = authenticate(username='myusername', password='supersecret', verify=False)\n",
```
bilts, line 50: This example no longer works as of the latest changes

harmony_py/auth.py
```diff
@@ -0,0 +1,146 @@
...
+        if self.auth is None:
+            # .netrc might have more auth for us on our new host.
+            new_auth = get_netrc_auth(prepared_request.url) if self.trust_env else None
+            if new_auth is not None:
+                prepared_request.prepare_auth(new_auth)
+
+        return
+
+
+def authenticate(*, auth: Optional[Union[Tuple[str, str], None]] = None,
```
bilts, line 92: This is new syntax to me.  What is `*` as the first param?

Also, did we end up with:
`authenticate(auth=(username, password))`

Which could also be shortened to:
`authenticate((username, password))`

I'm curious about the reasoning.

bilts, line 92: (Nit) Given that this method needs to be called by anyone using the library and is the entrypoint to all of the rest of the library:

1. I'd strongly consider moving / mirroring it into __init__.py so that the root package makes it available
2. I'd more weakly consider calling it something like "create_client" (I think we're using the word "session" elsewhere, "create_session" may work as well, though I'm not sure if it implies what we want)

harmony_py/auth.py
```diff
@@ -0,0 +1,146 @@
...
+                 validate: Optional[bool] = True) -> Session:
+    """
+    Create a requests-futures session for authenticated HTTP calls after optionally validating
```
bilts, line 95: What's the expected use of this session?

I was thinking (as alluded to in prior comments) that it'd be something like:
```
harmony_client = harmony.authenticate(...)
harmony_client.subset(...)
```
but is it actually
```
session = harmony.authenticate(...)
harmony.subset(session, ...)
```

examples/authentication.ipynb
```diff
@@ -0,0 +1,70 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# note: to run, you may want to create a soft link to the library for now: `ln -s ../harmony_py/ harmony_py`\n",
+    "# note: the keyword argument `verify` defaults to True, meaning it will verify the credentials are valid during session instantiation.\n",
+    "\n",
+    "from harmony_py.auth import authenticate\n",
```
bilts, line 47: Do we want `from harmony_py` or `from harmony`?

Note: potential name clash with https://pypi.org/project/harmony-pytorch/

bilts (changes requested):

harmony_py/auth.py
```diff
@@ -0,0 +1,197 @@
...
+    Returns:
+        (:obj:`SessionWithHeaderRedirection`): The authenticated requests session.
+
+    :raises MissingCredentials: No credentials were specified or found.
+    """
+
+    if username and password:
+        return SessionWithHeaderRedirection(username, password)
+    elif username and not password:
+        password = getpass()
```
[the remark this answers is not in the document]
rwat: harmony-py only uses authentication in conjunction with connections. We could separate them but one will always call the other.

harmony_py/auth.py
kbeamnsidc: Essentially it forces the remaining arguments to be keyword-only. 

Using a keyword to pass the tuple we thought was preferable since we will add the ability to pass a token later. It communicates in the type signature that a tuple (username,password) can be passed, and (later) that a token can be passed. This is also how requests works (`auth=(username,pwd)`). Maybe instead of `auth` this should be named `creds` or something similar.

By the way I think this type signature is wrong. Shouldn't it be `Optional[Tuple[str, str]] = None`? 

Later when we add a token I think it would mean adding: `token: Optional[str]`.

In the end any of these will be valid (let's pretend we change `auth` to `creds`):

```
authenticate()  # envvars or netrc file
authenticate(creds=('kbeam', 'secret'))
authenticate(token='xyzzy')
```

examples/authentication.ipynb
kbeamnsidc: I think I'd prefer we name the package `harmony` too.

harmony_py/auth.py
rwat: Makes sense - I'll plan to do option #1.

harmony_py/auth.py
rwat: That makes sense - it should be more of a 'client' (your first example).

harmony_py/auth.py
rwat: Changing auth to creds.

rwat: > By the way I think this type signature is wrong. Shouldn't it be Optional[Tuple[str, str]] = None?

I swear I had that before and something complained. It's working now though - changed.

tests/test_auth.py
```diff
@@ -0,0 +1,109 @@
...
+    response_mock = mocker.PropertyMock()
+    response_mock.request.configure_mock(url='https://www.othersite.gov')
+
+    mocker.patch('harmony_py.auth.PreparedRequest', return_value=preparedrequest_mock)
+    mocker.patch('harmony_py.auth.Response', return_value=response_mock)
+
+    session_with_creds = SessionWithHeaderRedirection(username='foo', password='bar')
+    session_with_creds.rebuild_auth(preparedrequest_mock, response_mock)
+
+    assert preparedrequest_mock.url == preparedrequest_props['url'] \
```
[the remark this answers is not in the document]
rwat: Either way they have to be built - Response, more so.

harmony_py/auth.py
```diff
@@ -0,0 +1,197 @@
...
+    1) If ``username`` and ``password`` are not None, create a session.
+    2) If ``username`` is specified but not ``password``, prompt user for a password and create a
+    session.
+    3) If ``netrc_file`` is True, rely on the automatic behavior of requests to find relevant
+    credentials in a .netrc file.
+    4) Attempt to read a username and password from environment variables, either from the system
+    or from a .env file to return a session.
+
+    Parameters:
+        username (str, optional): The EDL username.
```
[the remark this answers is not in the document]
kbeamnsidc: I'll get sphinx docs working for python code in 706

examples/authentication.ipynb
rwat: This is now done

bilts (approved):
👍

PR #16, opened 2021-03-11
Harmony 709: receive status messages from Harmony after submitting data requests

rwat:
Added status() method to Client, submit() now returns job_id, and various other changes.

### DIFF
--- a/harmony/harmony.py
+++ b/harmony/harmony.py
@@ -161,7 +161,6 @@ class Client:
             should_validate_auth: Whether EDL credentials will be validated.
         """
         self.config = Config(env)
-        self.hostname: str = self.config.hostname
         self.session = None
         self.auth = auth
 
@@ -176,7 +175,7 @@ class Client:
     def _url(self, request: Request) -> str:
         """Constructs the URL from the given request."""
         return (
-            f'https://{self.hostname}/{request.collection.id}'
+            f'https://{self.config.harmony_hostname}/{request.collection.id}'
             '/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset'
         )
 
--- a/harmony/harmony.py
+++ b/harmony/harmony.py
@@ -1,5 +1,7 @@
 from typing import NamedTuple
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Tuple
+
+import dateutil.parser
 
 from harmony.auth import create_session, validate_auth
 from harmony.config import Config, Environment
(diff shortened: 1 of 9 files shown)


harmony/harmony.py
```diff
@@ -209,18 +211,49 @@ def _temporal_subset_params(self, request: Request) -> list:
...
+        can't be reached.
+        """
+        session = self._session()
+        response = session.get(self._status_url(job_id)).result()
+        if response.ok:
+            if progress_only:
+                return (response.json())['progress']
+            else:
+                fields = [
+                    'status', 'message', 'progress', 'createdAt', 'updatedAt', 'request',
```
kbeamnsidc, line 253: Should we convert str -> datetime for createdAt and updatedAt?

harmony/harmony.py
```diff
@@ -209,18 +211,49 @@ def _temporal_subset_params(self, request: Request) -> list:
...
+        :raises Exception: This can happen if an invalid job_id is provided or Harmony services
+        can't be reached.
+        """
+        session = self._session()
+        response = session.get(self._status_url(job_id)).result()
+        if response.ok:
+            if progress_only:
+                return (response.json())['progress']
+            else:
+                fields = [
```
kbeamnsidc, line 252: We might want to python-ize the names (camel -> snake case)

kbeamnsidc (changes requested):

harmony/harmony.py
```diff
@@ -209,18 +211,49 @@ def _temporal_subset_params(self, request: Request) -> list:
...
         if response.ok:
-            job = response.json()
+            job_id = (response.json())['jobID']
         else:
             response.raise_for_status()
 
-        return job
+        return job_id
+
+    def status(self, job_id: str, progress_only=False) -> Union[str, dict]:
```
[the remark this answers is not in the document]
rwat: Will do.

harmony/harmony.py
rwat: I can do that.

harmony/harmony.py
rwat: Sure - as we've been making harmony-py it's sometimes hard to know how much of Harmony should be represented as-is vs. re-interpreting output on behalf of the user.

harmony/harmony.py
rwat: Note that I've chosen to import another third party library to handle ISO 8601 datetimes:

https://stackoverflow.com/questions/969285/how-do-i-translate-an-iso-8601-datetime-string-into-a-python-datetime-object
"As from python 3.7 you can use datetime.datetime.fromisoformat" but "The official documentation from the link you gave reads the following caution: This does not support parsing arbitrary ISO 8601 strings - it is only intended as the inverse operation of datetime.isoformat(). A more full-featured ISO 8601 parser, dateutil.parser.isoparse is available in the third-party package dateutil. So yeah, even for Python 3.7 we are back to the dateutil package."

I verified datetime.datetime.fromisoformat() is unable to parse an example Harmony-generated datetime '2021-02-25T19:18:39.798Z'. I was able to parse the datetime string with:

import dateutil.parser
dateutil.parser.parse('2021-02-25T19:18:39.798Z')

rwat: Hmm - looks like that was already satisfied by another import but I added it to `requirements/core.txt`.

kbeamnsidc (approved):

PR #17, opened 2021-03-12
Harmony-754: Add support for UMM-Var variables

kbeamnsidc:


rwat (commented):
The one thing I would mention is that I believe we're planning on having all harmony-py requests be async but I can work that in to harmony-710. Right now I'm assuming you'll merge harmony-709 into 708 and 754 and then I'll be merging the lot into 710 (ready soon(tm)).

PR #20, opened 2021-03-17
HARMONY-710: Job results and file downloading

rwat:
WIP

This PR is for initial review on the approach taken for blocking and non-blocking job methods. Due to limitations with `requests-futures` I continue to use futures for eventual results but rolled in a different file downloading setup. The initial design tried to hide any use of futures from the end user but it ended up being bad, I think, because it removed the ability for the user to cancel downloads early. The current design is blocking on a per download basis but continues to download all other files in parallel up to the thread pool worker size. That means someone may work with / process a file while the others are streaming to disk. Other features present.

Please see examples/job_results.ipynb

### DIFF
--- a/harmony/config.py
+++ b/harmony/config.py
@@ -28,7 +28,8 @@ class Config:
     """
 
     config = {
-        'NUM_REQUESTS_WORKERS': '8',
+        'NUM_REQUESTS_WORKERS': '3',  # increase for servers
+        'DOWNLOAD_CHUNK_SIZE': str(4 * 1024 * 1024)  # recommend 16MB for servers
     }
 
     def __init__(self, environment: Environment = Environment.UAT) -> None:
--- a/harmony/harmony.py
+++ b/harmony/harmony.py
@@ -385,8 +385,8 @@ class Client:
 
     def wait_for_processing(self, job_id: str, show_progress=False) -> None:
         check_interval = 3.0  # in seconds
-        sleep_interval = 0.33  # in seconds
-        intervals = int(check_interval / sleep_interval)
+        ui_update_interval = 0.33  # in seconds
+        intervals = int(check_interval / ui_update_interval)
         if show_progress:
             with progressbar.ProgressBar(max_value=100, widgets=progressbar_widgets) as bar:
                 progress = None
@@ -399,7 +399,7 @@ class Client:
                     if progress >= 100:
                         break
                     else:
-                        time.sleep(sleep_interval)
+                        time.sleep(ui_update_interval)
         else:
             while self.progress(job_id) < 100:
                 time.sleep(check_interval)
--- a/harmony/harmony.py
+++ b/harmony/harmony.py
@@ -382,7 +382,18 @@ class Client:
         else:
             response.raise_for_status()
 
-    def wait_for_processing(self, job_id: str, show_progress=False) -> None:
+    def wait_for_processing(self, job_id: str, show_progress: bool = False) -> None:
+        """Retrieve a submitted job's completion status in percent.
+
+        Args:
+            job_id: UUID string for the job you wish to interrogate.
+
+        Returns:
+            The job's processing progress as a percentage.
+
+        :raises Exception: This can happen if an invalid job_id is provided or Harmony services
+        can't be reached.
+        """
         check_interval = 3.0  # in seconds
         ui_update_interval = 0.33  # in seconds
         intervals = int(check_interval / ui_update_interval)
@@ -403,17 +414,40 @@ class Client:
             while self.progress(job_id) < 100:
                 time.sleep(check_interval)
 
-    def result_json(self, job_id: str, show_progress=False) -> str:
+    def result_json(self, job_id: str, show_progress: bool = False) -> str:
+        """Retrieve a job's final json output.
+
+        Harmony jobs' output is built as the job is processed and this method fetches the complete
+        and final output.
+
+        Args:
+            job_id: UUID string for the job you wish to interrogate.
+            show_progress: Whether a progress bar should show via stdout.
+
+        Returns:
+            The job's complete json output.
+        """
         self.wait_for_processing(job_id, show_progress)
         response = self._session().get(self._status_url(job_id))
         return response.json()
 
-    def result_urls(self, job_id: str, show_progress=False) -> List:
+    def result_urls(self, job_id: str, show_progress: bool = False) -> List:
+        """Retrieve the data URLs for a job.
+
+        The URLs include links to all of the jobs data output.
+
+        Args:
+            job_id: UUID string for the job you wish to interrogate.
+            show_progress: Whether a progress bar should show via stdout.
+
+        Returns:
+            The job's complete list of data URLs.
+        """
         data = self.result_json(job_id, show_progress)
         urls = [x['href'] for x in data['links'] if x['rel'] == 'data']
         return urls
 
-    def _download_file(self, url, directory=None, overwrite=False) -> str:
+    def _download_file(self, url: str, directory: str = '', overwrite: bool = False) -> str:
         chunksize = int(self.config.DOWNLOAD_CHUNK_SIZE)
         session = self._session()
         filename = url.split('/')[-1]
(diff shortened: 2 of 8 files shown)


harmony/harmony.py
```diff
@@ -260,7 +272,7 @@ def _status_url(self, job_id: str) -> str:
 
     def _params(self, request: Request) -> dict:
         """Creates a dictionary of request query parameters from the given request."""
-        params = {}
+        params = {'forceAsync': True}
```
chris-durbin, line 275: Looks like this means users cannot override to allow sync requests - is that true, and is that what we want?

harmony/harmony.py
```diff
@@ -364,8 +376,73 @@ def progress(self, job_id: str) -> int:
...
-        response = session.get(self._status_url(job_id)).result()
+        response = session.get(self._status_url(job_id))
         if response.ok:
             return int((response.json())['progress'])
         else:
             response.raise_for_status()
+
+
+    def wait_for_processing(self, job_id: str, show_progress=False) -> None:
+        check_interval = 3.0  # in seconds
```
chris-durbin, line 387: Do we want to make these values overridable?

I'm also trying to understand the difference between the check_interval and the sleep interval. Couldn't we just have one interval? It looks like on line 389 intervals is set to 9 seconds and then on 405 we sleep for 0.33 seconds before checking again. I guess I'm a little confused by how long we're sleeping and how often we're calling harmony.

harmony/config.py
```diff
@@ -28,7 +28,8 @@ class Config:
     """
 
     config = {
-        'NUM_REQUESTS_WORKERS': '8',
+        'NUM_REQUESTS_WORKERS': '3',  # increase for servers
```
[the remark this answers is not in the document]
rwat: I was thinking for machine-to-machine use. So, the initial and common use-case are individual end-users and the default settings were set with them in mind. However, if software projects like AppEEARS want to use `harmony-py` then they would want different 'server' settings. Yeah, I didn't make that very clear in the comment.

harmony/harmony.py
rwat: That is true. They can get sync-like behavior by submitting the request followed by running any of the following blocking method:
- wait_for_processing()
- result_json()
- result_urls()

It'd also be possible to alter the behavior of submit() to optionally block right away. What I was trying to do is present a sync or async experience through `harmony-py` while at the same time as having a consistent back-end workflow between `harmony-py` and Harmony.

harmony/harmony.py
rwat: `check_interval` is how often we're calling Harmony and is meant to be set to an interval that won't produce undue load on the servers. The sleep interval determines how often we refresh the UI (the optional progress bar gui element). Dividing the two gives us a number we can use to create the cycle'd sequence the loop goes through so that it updates the UI fast but checks the servers slowly.

rwat: Changing `sleep_interval` to `ui_update_interval`.

chris-durbin (approved):

rwat:
Ready for final / full review.

kbeamnsidc (approved):

harmony/harmony.py
bilts: Even if not overridable, putting it in some sort of constant and moving the various constants to a consistent location (top of the file?) is a good practice.  I think Chris may be responding to a sense of inconsistency and magic in seeing a number but not knowing what it's for or where it came from.  Rule of thumb: if the number isn't 0 or 1, put it in a constant and add a line of docs.  I patiently await you finding 5 places I've personally broken that rule :)

bilts: (Looking through the code) Also numbers that are obviously well-understood and will basically never ever change, like checking if an HTTP status is 404, you don't need to externalize that.

chris-durbin: Just FYI that Ryan and I paired to walk through a notebook using this logic and renaming sleep_interval to ui_update_interval made a big difference to grok what the code was doing.

I do agree that pulling out constants is good - defining them here in the function is ok for me too.

rwat: Good feedback all around. I moved those constants up to the top of harmony.py and commented on their purpose.

bilts:
I'm seeing it saying "Processing: N/A%" initially in the notebook

harmony/harmony.py
```diff
@@ -364,8 +376,150 @@ def progress(self, job_id: str) -> int:
...
+        if show_progress:
+            with progressbar.ProgressBar(max_value=100, widgets=progressbar_widgets) as bar:
+                progress = None
+                check_cycle = cycle([True] + list(repeat(False, intervals - 1)))
+                for should_get in check_cycle:
+                    if should_get:
+                        progress = self.progress(job_id)
+                    bar.update(progress)
+                    sys.stdout.flush()  # ensures correct behavior in Jupyter notebooks
+                    if progress >= 100:
```
bilts, line 412: What if there's an error in the request, so progress < 100, but the request is done?

harmony/harmony.py
```diff
@@ -364,8 +376,150 @@ def progress(self, job_id: str) -> int:
...
+        :raises Exception: This can happen if an invalid job_id is provided or Harmony services
+        can't be reached.
+        """
+        check_interval = 3.0  # in seconds
+        ui_update_interval = 0.33  # in seconds
+        intervals = int(check_interval / ui_update_interval)
+        if show_progress:
+            with progressbar.ProgressBar(max_value=100, widgets=progressbar_widgets) as bar:
+                progress = None
+                check_cycle = cycle([True] + list(repeat(False, intervals - 1)))
```
bilts, line 406: I could use some explanation on this one (maybe a comment).  Why does it only check every nth time around the loop?

harmony/harmony.py
```diff
@@ -364,8 +376,150 @@ def progress(self, job_id: str) -> int:
...
+                    if should_get:
+                        progress = self.progress(job_id)
+                    bar.update(progress)
+                    sys.stdout.flush()  # ensures correct behavior in Jupyter notebooks
+                    if progress >= 100:
+                        break
+                    else:
+                        time.sleep(ui_update_interval)
+        else:
+            while self.progress(job_id) < 100:
```
bilts, line 422: Same comment.  I don't think this will work if there's an error processing the job.

harmony/harmony.py
bilts: Rereading, I'm pretty sure I get it, but I think it could be more clear.  Maybe something like:

```python
progress = self.progress(job_id)
# Update the UI `intervals` times for each status fetch
for i in range(intervals):
   bar.update(progress)
   sys.stdout.flush()
   if progress >= 100: # or an error. see other comments
       break
   else:
       time.sleep(ui_update_interval)
```

harmony/harmony.py
```diff
@@ -364,8 +376,150 @@ def progress(self, job_id: str) -> int:
...
+        ui_update_interval = 0.33  # in seconds
+        intervals = int(check_interval / ui_update_interval)
+        if show_progress:
+            with progressbar.ProgressBar(max_value=100, widgets=progressbar_widgets) as bar:
+                progress = None
+                check_cycle = cycle([True] + list(repeat(False, intervals - 1)))
+                for should_get in check_cycle:
+                    if should_get:
+                        progress = self.progress(job_id)
+                    bar.update(progress)
```
bilts, line 415: I'm curious what this code is actually meant to do.  `progress` can't change values except when line 406 gets called, right?

harmony/harmony.py
bilts: You can ignore my prior comments, I think what you have is probably alright.

harmony/harmony.py
rwat: ProgressBar has a "spinner" UI element that changes on each update. There are two update intervals we want to work with: how often to fetch results from the server and how often to update the UI. We could make them the same interval but the spinner wouldn't move very much. It has a better aesthetic when it looks like it's more active or 'working more'. So I went with the smaller interval of the two and paired that with a sequence (cycle) of booleans. On True, do both UI updating and server checking. On False, only do UI updating. There's certainly other ways to do this sort of thing like keep track of timestamps and do server updates after a threshold period of time; reset threshold start time and do again, etc.

harmony/harmony.py
rwat: Then this code is stuck  : (  Good point.

harmony/harmony.py
rwat: I think we wrote the last two comments at the same time  : ) Sure - I'll get that updated.

harmony/harmony.py
rwat: It's the spinner - see above comment.

harmony/harmony.py
rwat: This code has now been updated to respond to failed and canceled jobs.

harmony/harmony.py
rwat: Updated code to be cleaner / simpler.

harmony/harmony.py
rwat: Code has been updated.

harmony/harmony.py
```diff
@@ -364,8 +376,171 @@ def progress(self, job_id: str) -> int:
...
+        intervals = round(check_interval / ui_update_interval)
+        if show_progress:
+            with progressbar.ProgressBar(max_value=100, widgets=progressbar_widgets) as bar:
+                progress = 0
+                while progress < 100:
+                    progress, status = self.progress(job_id)
+                    if status == 'failed':
+                        raise Exception('Job has failed. Call result_json() to learn more.')
+                        break
+                    if status == 'canceled':
```
bilts, line 411: no string localization? ;)

bilts (approved):

PR #21, opened 2021-03-23
HARMONY-711: Added job STAC URL retrieval in harmony.py.

rwat:
HARMONY-711: Added arbitrary URL retrieval from Client to support pystac
common usage.
HARMONY-711: Added example STAC notebook job_stac.ipynb.
HARMONY-711: examples.txt requirements file now uses pystac.

This card was just a small effort to retrieve the STAC URL and make it possible to feed STAC data into something like pystac.

The review notebook is in examples/job_stac.ipynb

### DIFF
--- a/examples/job_stac.ipynb
+++ b/examples/job_stac.ipynb
@@ -121,7 +121,11 @@
    "source": [
     "from pystac import Catalog\n",
     "\n",
-    "cat = Catalog.from_file(stac_catalog_url)"
+    "cat = Catalog.from_file(stac_catalog_url)\n",
+    "\n",
+    "print(cat.title)\n",
+    "for item in cat.get_all_items():\n",
+    "    print(item.datetime, [asset.href for asset in item.assets.values()])"
    ]
   },
   {
--- a/examples/job_stac.ipynb
+++ b/examples/job_stac.ipynb
@@ -1,0 +1,135 @@
+{
+ "metadata": {
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.8.5-final"
+  },
+  "orig_nbformat": 2,
+  "kernelspec": {
+   "name": "python3",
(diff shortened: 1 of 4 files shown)


kbeamnsidc (approved):

harmony/harmony.py
```diff
@@ -539,8 +540,47 @@ def download_all(self,
             The filename and path.
         """
         urls = self.result_urls(job_id, show_progress=False) or []
-        file_names = []
+        futures = []
         for url in urls:
             future = self.executor.submit(self._download_file, url, directory, overwrite)
```
bilts, line 545: nit (and not code you touched in this PR, and doesn't need to be fixed, but while we're here)... this whole method could and probably should be:
```python
return [self.download(url, directory, overwrite) for url in self.result_urls(job_id)]
```

examples/job_stac.ipynb
```diff
@@ -0,0 +1,135 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from pystac import Catalog\n",
+    "\n",
+    "cat = Catalog.from_file(stac_catalog_url)"
```
bilts, line 124: This notebook doesn't have enough output to really show that it worked.  Try the following in a cell:

```python
print(cat.title)
for item in cat.get_all_items():
    print(item.datetime, [asset.href for asset in item.assets.values()])
```

bilts:
Will approve if at least the comment about notebook output is addressed.

harmony/harmony.py
rwat: That's cool. I've occasionally introduced a variable in places like that to assist in debugging but I agree it's nicer not to populate the namespace any more than is necessary.

examples/job_stac.ipynb
rwat: I think you wrote more of this card than I did  : ) I definitely appreciate the help, though.

kbeamnsidc (approved):

harmony/harmony.py
bilts: Sure, extracting into a var would be alright, too.  I think the important piece is to avoid repeating the body of `download` inside of `download_all` so that if we change how downloading works we don't have to remember to do it in 2 places.

bilts (approved):

PR #23, opened 2021-03-25
Harmony-703: Enabling packaging of the library for PyPI

rwat:
The branch includes all the files and changes needed to support packaging and uploading to PyPI in order to enable users of harmony-py to 'pip install'.


chris-durbin:
I left a comment on the wiki page. It is saying there's a test failure for 3.7, but it looks like it failed once, but on PR that worked ok. I'm not sure what's going on there.

setup.py
```diff
@@ -0,0 +1,209 @@
...
+    # List additional groups of dependencies here (e.g. development
+    # dependencies). Users will be able to install these using the "extras"
+    # syntax, for example:
+    #
+    #   $ pip install sampleproject[dev]
+    #
+    # Similar to `install_requires` above, these must be valid existing
+    # projects.
+    #extras_require={  # Optional
+    #    'dev': ['check-manifest'],
```
bilts, line 164: May want this.  It could help greatly with notebook execution, i.e. `pip install -e .[dev]`

setup.py
```diff
@@ -0,0 +1,209 @@
...
+    #    'test': ['coverage'],
+    #},
+
+    # If there are data files included in your packages that need to be
+    # installed, specify them here.
+    #package_data={  # Optional
+    #    'sample': ['package_data.dat'],
+    #},
+
+    # Although 'package_data' is the preferred approach, in some case you may
```
bilts, line 174: Iffy on whether we should include all the commented out stuff if we don't plan to use it.  Would probably err on the side of nuking it.

bilts:
Consider adding `make build` and `make publish` targets, e.g. https://github.com/nasa/harmony-service-lib-py/blob/main/makefile#L11

setup.py
```diff
@@ -0,0 +1,209 @@
...
+    python_requires='>=3.6, <4',
+
+    # This field lists other packages that your project depends on to run.
+    # Any package you put here will be installed by pip when your project is
+    # installed, so they must be valid existing projects.
+    #
+    # For an analysis of "install_requires" vs pip's requirements files see:
+    # https://packaging.python.org/en/latest/requirements.html
+    install_requires=[
+        'python-dateutil', 'python-dotenv', 'progressbar2', 'requests', 'sphinxcontrib-napoleon'
```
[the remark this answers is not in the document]
rwat: I can read from the requirements files instead. I definitely see the advantage of having version requirements in one place.

Regarding the decision not to put specific versions, I was following the recommendation here:
https://packaging.python.org/discussions/install-requires-vs-requirements/#install-requires-vs-requirements-files

> It is not considered best practice to use install_requires to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades.

> Lastly, it’s important to understand that install_requires is a listing of “Abstract” requirements, i.e just names and version restrictions that don’t determine where the dependencies will be fulfilled from (i.e. from what index or source). The where (i.e. how they are to be made “Concrete”) is to be determined at install time using pip options

rwat:
@chris-durbin Yeah, github has been failing on Actions sometimes. I'll run it a couple more times to clear it up but it'd be nice to know the root cause.

setup.py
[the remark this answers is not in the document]
bilts: I read that recommendation differently, that we don't want to pin to a specific `major.minor.patch` version, but not that you wouldn't want to specify a compatible range of versions.  So, not `3.2.1` but perhaps `~=3.2` to avoid issues if a user is using `1.x.x` or if `4.x.x` comes out with a breaking API.  This is how Poetry and NPM both work, where you specify a broad range of versions in your main package file, then they maintain a separate lock file that pins specific versions.

Note: I'm not sure I'm right, just reading it differently.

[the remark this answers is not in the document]
rwat: I was interpreting it like you said - any versions, but it makes a lot more sense to specify e.g. major.minor. I'll get those requirements files included.

rwat:
@bilts I completely changed setup.py to be similar to the harmony service library but I wasn't able to see why 'version' was treated the way it was there so I left in the VERSION file. It seemed like that would be the simplest option. Of course that's a deviation from the other library so if you'd still like to remove the use of that file then I'd like to have a quick conversation about it so I understand the importance of doing it that way.

Makefile
```diff
@@ -45,3 +49,13 @@ ci: lint test
 
 docs:
 	cd docs && $(MAKE) html
+
+build:
```
kbeamnsidc, line 53: Duplicate target?

kbeamnsidc:
Some meta-review comments:

- The ticket AC says to document the release process on the wiki, so this work follows the AC...but...I'd rather see that documented in the README or another doc *in* the project. If the versioning / release process changes, changing the docs should be in a commit. Perhaps the AC can be satisfied by moving the docs into the README in this repo (or another file), and rendering the README somewhere on the wiki.
- The wiki shows individual commands running, e.g., twine, but why doesn't it show `make publish` instead?
- In the ticket, @rwat you added a comment about releasing being reserved for POs. Is that true in general--only POs will do releases?
- Consider removing the VERSION file and instead creating a `harmony` package-level `__version__` variable. This makes it available in code and also simplifies the manifest. https://packaging.python.org/guides/single-sourcing-package-version/
- Consider moving from dynamic to static metadata, i.e., changing from `setup.py` to `setup.cfg`. See: https://packaging.python.org/tutorials/packaging-projects/#configuring-metadata
- Add a `pyproject.toml` since this is the standard for Python project metadata. See: https://packaging.python.org/tutorials/packaging-projects/#creating-pyproject-toml
- Versioning & releasing should be simpler and not require updating a value in a file. I.e., `poetry version minor` (but I'm not suggesting poetry, just using it as an example...see below).
- Versioning should also create a git tag on the repo with the version #.
- We should create a github release when we create a new version.

I'd like to see the release process be _at most_:
```
$ make version type=minor    # note this _is_ different than passing an envvar by setting it before the make invocation
$ make build
$ make publish
```

Ideally:
`$ make release type=minor`

bilts:
> @bilts I completely changed setup.py to be similar to the harmony service library but I wasn't able to see why 'version' was treated the way it was there so I left in the VERSION file. It seemed like that would be the simplest option. Of course that's a deviation from the other library so if you'd still like to remove the use of that file then I'd like to have a quick conversation about it so I understand the importance of doing it that way.

@rwat I think Kevin covered it well.  `__version__` makes the value readily available in code, which can be useful to clients supporting multiple versions or wishing to add debugging information

rwat:
@kbeamnsidc - thank you for great suggestions. Here's what I've changed:

- I deleted the Wiki page (which had incorrect information because I had failed to update it). The build and release information is now in README.md.
- I removed the verbage about PO's being the only ones who can release new versions of harmony-py though I think that's still the case. Not sure if that will change.
- I removed the VERSION file and replaced it with the same workflow harmony-service-lib-py uses (see README.md). Once the master branch is tagged with a new version, that information will cascade to all the necessary files and commands. So, git tag becomes the single source of version truth for the project.
- regarding setup.cfg vs. setup.py and static vs. dynamic data, setup.py is currently 66 lines including comments. It's fast to review and as long as it remains small I think there's a win to not increase file count, etc. (complexity). I'd like to keep setup.py as-is unless I'm getting overruled  : )
- I added a pyproject.toml file.
- I think versioning and releasing is simpler now; it's possible to reduce the work to two commands: one for tagging, one for publishing.
- I asked the group about github releases and it sounds like we'll do those manually for now.

kbeamnsidc (approved):

kbeamnsidc (approved):

README.md
```diff
@@ -116,3 +116,22 @@ Either set `HARMONY_DIR` or replace it with your Harmony project directory path.
...
+
+Then, to generate new package and wheel files:
+
+        $ make build
+
+The Makefile reads the current version number based on git tag, populates the version in `harmony/__init__.py`, and `setup.py` reads the version number from `harmony/__init__.py` for packaging purposes.
+
+Provided API tokens are in order, the following runs the build target and publishes to PyPI:
+
+        # make publish
```
bilts, line 135: If I understand correctly, following these steps will:

- Tag the current branch locally
- Update the contents of `harmony\__init__.py` locally
- Build the package / wheel
- Publish to PyPI

At the end of doing that, there will be an uncommitted change with the correct contents of `harmony/__init__.py` and a _local_ tag with the correct version but not including the change to `harmony/__init__.py`

Thoughts about having `make build` commit the version and `git tag -f` (tag or move an existing tag) the version?  I think keeping it as either a manual git push or putting the push in `make publish` is best so the scary stuff happens in one place.

README.md
```diff
@@ -116,3 +116,22 @@ Either set `HARMONY_DIR` or replace it with your Harmony project directory path.
 Harmony-py uses [GitHub
 Actions](https://github.com/nasa/harmony-py/actions) to run the Linter
 & Unit Tests. The test coverage output is saved as a build artifact.
+
+## Building and Releasing
+
+If a new version of Harmony-Py has been released then `master` should be tagged with an updated version:
+
+        $ git checkout master
+        $ git tag -a 1.2.3    # where 1.2.3 is the next version number
```
bilts, line 125: harmony-service-lib-py prefixes its version tags with `v`.  Can we be consistent with that?

README.md
bilts: Alternatively, is your thinking that we'd never commit `__init__.py` and always have it some default value in copies from git?  I think that could work, too.

bilts (commented):
Very close.  I think the following are needed before approval:

* Update so this works with a prefix of `v` on our version tags, for consistency
* Answer review comments about whether the version in `__init__.py` is to be committed and update code if so

I plan to test this by actually doing the push once we merge.

README.md
kbeamnsidc: There's another I've heard about called versioneer (https://github.com/python-versioneer/python-versioneer) but I've not used it. I think these things are "on the outs" since tools like poetry do this.

rwat: Based on recent conversations, I guess I would propose the following manual steps and revisit / finish automating in a future ticket. Presumably that ticket would implement a method of versioning and building across all GH hosted projects (or all/most python projects).

In summary:
- `git tag` determines version information
- `make build` stores version info in __init __.py based on tag
- we would commit that change to master, re-apply the tag, then push to master
- `make publish`

kbeamnsidc (approved):

bilts (approved):

PR #25, opened 2021-04-01
Harmony-715: Add support for requesting s3 URLs

kbeamnsidc:
Note that this is branched off of 714 and shows a diff from that branch for now (will merge into master after 714).

This PR also includes extending the Config to support localhost URLs for testing with Harmony running locally.


harmony/config.py
```diff
@@ -15,9 +15,10 @@
 
 from dotenv import load_dotenv
 
-Environment = Enum('Environment', ['SBX', 'SIT', 'UAT', 'PROD'])
+Environment = Enum('Environment', ['LOCAL', 'SBX', 'SIT', 'UAT', 'PROD'])
 
 HOSTNAMES = {
+    Environment.LOCAL: 'localhost:3000',
     Environment.SBX: 'harmony.sbx.earthdata.nasa.gov',
```
chris-durbin, line 22: We don't have an external URL for sandbox. We can only tunnel in and hit an internal load balancer.

harmony/config.py
```diff
@@ -15,9 +15,10 @@
 
 from dotenv import load_dotenv
 
-Environment = Enum('Environment', ['SBX', 'SIT', 'UAT', 'PROD'])
+Environment = Enum('Environment', ['LOCAL', 'SBX', 'SIT', 'UAT', 'PROD'])
 
 HOSTNAMES = {
+    Environment.LOCAL: 'localhost:3000',
```
chris-durbin, line 21: Is this something we can override? For example I tend to run my harmony on port 5000 locally.

harmony/config.py
```diff
@@ -55,12 +56,20 @@ def harmony_hostname(self):
         """Returns the hostname for this Config object's Environment."""
         return HOSTNAMES[self.environment]
 
+    @property
+    def url_scheme(self) -> str:
+        return 'http' if self.environment == Environment.LOCAL else 'https'
```
chris-durbin, line 61: Sandbox using an internal load balancer only has an http listener (so sandbox and local should both be http).

harmony/config.py
kbeamnsidc: I was going to make it configurable and then thought YAGNI, but with the sneaking suspicion that we would need it. Yep, I'll add this.

rwat: I run Harmony on a different port too.

harmony/config.py
kbeamnsidc: Should I just remove this? I.e., would we potentially expose sandbox in the future?

rwat:
Review partially done - I'd like to hold off until the decision over SBX is resolved and finish the review after any code changes related to that.

harmony/config.py
chris-durbin: Sandbox accounts aren't allowed to expose any http services. I don't think we need to make harmony-py work for sandbox, so removing anything referencing sandbox makes to me. If we need to make sandbox work that's a new ticket anyway.

rwat (approved):
Apart from the one outstanding item Chris pointed out, looks good to me.

examples/s3_access.ipynb
```diff
@@ -82,24 +98,40 @@
     "\n",
     "import boto3\n",
     "\n",
-    "s3 = boto3.client('s3', **creds)\n",
-    "uri = 's3://harmony-uat-staging/public/harmony/netcdf-to-zarr/817a3e99-d53c-4169-b9f1-82cc947793be/2020_01_01_7f00ff_global.zarr'\n",
-    "bucket_name = uri.split('/')[2]\n",
-    "object_name = '/'.join(uri.split('/')[3:])\n",
-    "file_name = uri.split('/')[-1]\n",
+    "def parse_s3_url(url):\n",
```
bilts, line 101: I think I made a note of this on the other ticket that had S3 work, but rather than require every use to adapt the URL to what works with boto3, I think we should default to returning a tuple of (bucket_name, object_name) or something similarly easy to drop into boto3

harmony/harmony.py
```diff
@@ -269,16 +279,17 @@ def _submit_url(self, request: Request) -> str:
...
 
-    def _status_url(self, job_id: str) -> str:
+    def _status_url(self, job_id: str, link_type: LinkType = LinkType.https) -> str:
         """Constructs the URL for the Job that is used to get its status."""
-        return f'https://{self.config.harmony_hostname}/jobs/{job_id}'
+        return f'{self.config.root_url}/jobs/{job_id}?linktype={link_type.value}'
 
     def _cloud_access_url(self) -> str:
-        return f'https://{self.config.harmony_hostname}/cloud-access'
+        return f'{self.config.root_url}/cloud-access'
```
bilts, line 292: Funny enough I did this exact same thing in the 721 branch.  Happy for the merge conflict if I can test locally.

chris-durbin (approved):

PR #27, opened 2021-04-02
HARMONY-721: Implement shapefile subsetting

bilts:
Test using commands in the provided example notebook


harmony/harmony.py
```diff
@@ -180,6 +196,12 @@ def __init__(self,
             (lambda tr: tr['start'] < tr['stop'] if 'start' in tr and 'stop' in tr else True,
              'The temporal range\'s start must be earlier than its stop datetime.')
         ]
+        self.shape_validations = [
+            (lambda s: os.path.isfile(s), 'The provided shape path is not a file'),
+            (lambda s: s.split('.').pop().lower() in _shapefile_exts_to_mimes,
+             'The provided shape file is not a recognized type.  Valid file extensions: '
```
indiejames, line 202: I know this approach to validation is consistent with what is done for spatial/temporal validation, but it has the drawback of not being able to include the provided type in the error message. Error messages like "You didn't provide a valid X. Valid Xs are ..." have always been a source of annoyance for me. I prefer "You provided 'foo' which is not a valid X. Valid Xs are ...".

indiejames (approved):
Just the one small comment about error messages. Verified using example notebook.

harmony/harmony.py
```diff
@@ -275,13 +295,15 @@ def _cloud_access_url(self) -> str:
 
     def _params(self, request: Request) -> dict:
         """Creates a dictionary of request query parameters from the given request."""
-        params = {'forceAsync': True}
+        params = {'forceAsync': 'true'}
```
chris-durbin, line 298: It would be nice to use a boolean rather than a string for true/false. Any reason this had to change?

chris-durbin (approved):

rwat:
Code looks great. Tests, docs, lint done.

I'm getting an error message when I run the example notebook and I wonder if it's EULA-related.:
> {"username":"rmwaters","status":"failed","message":"Failed","progress":0,"createdAt":"2021-04-02T18:02:14.832Z","updatedAt":"2021-04-02T18:02:32.513Z","links":[{"title":"Job Status","href":"https://harmony.uat.earthdata.nasa.gov/jobs/3f7da3bd-2e58-42f3-ab88-4210316d0ab0","rel":"self","type":"application/json"}],"request":"https://harmony.uat.earthdata.nasa.gov/C1233629671-ASF/ogc-api-coverages/1.0.0/collections/Band1/coverage/rangeset?forceAsync=true","numInputGranules":1,"jobID":"3f7da3bd-2e58-42f3-ab88-4210316d0ab0"}

harmony/harmony.py
bilts: It's the responsibility of this method to get values into their serialized form.  A few lines down on line 309, you can see where we are explicitly changing passed booleans into lower-case string representations.  This change wasn't technically needed, but the inconsistency and serialization of the value as 'True' instead of 'true' cost me debugging time, so I figured I'd change it.

bilts:
@rwat Given that others can independently verify the notebook, that you're unable to get the underlying data without going through Harmony at all, and access to the actual data in question is orthogonal to this ticket, I'm inclined to merge.

rwat (approved):

PR #29, opened 2021-04-08
HARMONY-787: Fix pip package installation

bilts:
To test:
```
make build
tar -ztf dist/harmony-py-*
```
Verify that the output includes harmony/* and not tests/*

In a clean Python environment, you can 
`pip install dist/harmony-py-*` and/or `pip install -e .` in the harmony-py dir, go to another dir (e.g. /tmp) run `python` and `from harmony import Client` and verify no exception is thrown

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,6 @@ import pathlib
 import re
 from setuptools import find_packages, setup
 
-
 DEPENDENCIES = []
 with open('requirements/core.txt', 'r') as f:
     DEPENDENCIES = f.read().strip().split('\n')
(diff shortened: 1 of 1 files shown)


chris-durbin (approved):

rwat (approved):
I did the following:

- make build
- tar -ztf dist/harmony-py-* and verified library files were present (and not tests)
- created a new venv; activated venv
- cd dist; pip install wheel; pip install ./harmony_py-0.1.0+19.g445e304-py3-none-any.whl  (I may not have needed the `pip install wheel` but I guess I did it anyway)
- cd ~   (<- i.e. outside of the repo)
- python;; from harmony import Client

No errors!  Thanks Patrick.

PR #30, opened 2021-04-28
Harmony 801

hailiangzhang:
This PR:
1. adds user-agent headers when a request is submitted via Client
2. adds tests to make sure the user-agent headers are being added with the expected format.

### DIFF
--- a/harmony/harmony.py
+++ b/harmony/harmony.py
@@ -494,7 +494,6 @@ class Client:
         session = self._session()
         params = self._params(request)
         headers = self._headers()
-        print(headers)
 
         with self._files(request) as files:
             if files:
@@ -508,9 +507,10 @@ class Client:
                 file_items = [(k, v) for k, v in files.items()]
                 response = session.post(
                     self._submit_url(request),
-                    files=param_items + file_items)
+                    files=param_items + file_items,
+                    headers=headers)
             else:
-                response = session.get(self._submit_url(request), params=params)
+                response = session.get(self._submit_url(request), params=params, headers=headers)
         if response.ok:
             print(response.json())
             job_id = (response.json())['jobID']
--- a/harmony/harmony.py
+++ b/harmony/harmony.py
@@ -363,9 +363,52 @@ class Client:
         return params
 
     def _headers(self) -> dict:
-        """Create (if needed) and returns a dictionary of headers."""
-        if self.headers is None:
-            self.headers = {}
+        """
+        Create (if needed) and return a dictionary of headers.
+        Code partially adaped from: https://github.com/requests/toolbelt/blob/master/requests_toolbelt/utils/user_agent.py
+        """
+        if 'headers' not in self.__dict__:
+            user_agent_content = set([])
+
+            # Get harmony package info
+            try:
+                import harmony
+                user_agent_content.add(f'harmony-py/{harmony.__version__}')
+            except Exception as e:
+                print("Following exception was caught when building user-agent headers for harmony-py.")
+                print(e)
+
+            # Get platform info
+            try:
+                p_system = platform.system()
+                p_release = platform.release()
+                user_agent_content.add(f'{p_system}/{p_release}')
+            except Exception as e:
+                print("Following exception was caught when building user-agent headers for harmony-py.")
+                print(e)
+
+            # Get implementation info
+            implementation = platform.python_implementation()
+            if implementation == 'CPython':
+                implementation_version = platform.python_version()
+            elif implementation == 'PyPy':
+                implementation_version = '%s.%s.%s' % (sys.pypy_version_info.major,
+                                                       sys.pypy_version_info.minor,
+                                                       sys.pypy_version_info.micro)
+                if sys.pypy_version_info.releaselevel != 'final':
+                    implementation_version = ''.join([
+                        implementation_version, sys.pypy_version_info.releaselevel
+                        ])
+            elif implementation == 'Jython':
+                implementation_version = platform.python_version()  # Complete Guess
+            elif implementation == 'IronPython':
+                implementation_version = platform.python_version()  # Complete Guess
+            else:
+                print("Unable to collect python implementation info when building user-agent headers for harmony-py.")
+
+            # Build headers
+            self.headers = {'User-Agent': ' '.join(user_agent_content)}
+
         return self.headers
 
     def _spatial_subset_params(self, request: Request) -> list:
@@ -449,6 +492,8 @@ class Client:
         job_id = None
         session = self._session()
         params = self._params(request)
+        headers = self._headers()
+        print(headers)
 
         with self._files(request) as files:
             if files:
@@ -470,8 +515,6 @@ class Client:
             job_id = (response.json())['jobID']
         else:
             response.raise_for_status()
-
-        # Add user-agent headers
 
         return job_id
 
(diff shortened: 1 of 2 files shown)


hailiangzhang:
Looks like there is a CICD error for python3.7 only:
https://github.com/nasa/harmony-py/pull/30/checks?check_run_id=2459983961
image
Does this seem to be a CICD issue?

hailiangzhang:
Not sure what happened, but it looks like all builds passed now.

tests/test_client.py
```diff
@@ -263,6 +264,36 @@ def test_with_invalid_request():
...
 
 
+@responses.activate
+def test_request_has_user_agent_headers():
+    collection = Collection('foobar')
+    request = Request(
+        collection=collection,
+    )
+    responses.add(
+        responses.GET,
```
indiejames, line 274: You should probably add a case for POST requests as well.

harmony/harmony.py
```diff
@@ -361,6 +362,49 @@ def _params(self, request: Request) -> dict:
...
+        """
+        Create (if needed) and return a dictionary of headers.
+        Code partially adaped from:
+            https://github.com/requests/toolbelt/blob/master/requests_toolbelt/utils/user_agent.py
+        """
+        if 'headers' not in self.__dict__:
+            user_agent_content = set([])
+
+            # Get harmony package info
+            try:
```
indiejames, line 375: I think you can just use one `try` block for all three since the exception handling is identical.

harmony/harmony.py
```diff
@@ -361,6 +362,49 @@ def _params(self, request: Request) -> dict:
...
+                print(e)
+
+            # Get implementation info
+            try:
+                implementation = platform.python_implementation()
+                implementation_version = platform.python_version()
+                user_agent_content.add(f'{implementation}/{implementation_version}')
+            except Exception as e:
+                print("Following exception was caught "
+                      "when building user-agent headers for harmony-py:")
```
indiejames, line 400: I'm not sure it is a good idea to swallow the exceptions like this. In any case you probably don't want to return a possibly incorrect header on line 404.

indiejames (changes requested):
Looks pretty good, just a few comments.

tests/test_client.py
hailiangzhang: Yeah, I realized that after sending out PR ... if it's easy for you, could you point me to a minimal test case using POST?

harmony/harmony.py
hailiangzhang: Yes, will do.

hailiangzhang: Done.

hailiangzhang: Hmm, had a second thought, maybe we do want to keep separate `try` blocks, because I think it's Okay to miss some of the user-agent header items. If we merge them into one block, we may miss some header items to be generated after those throwing exceptions.

hailiangzhang: Reverted back to "separate try blocks".

harmony/harmony.py
hailiangzhang: Well, if an exception is caught, it will not be saved to the user agent header (`user_agent_content`).
But a good point is that, if all header items failed, it probably doesn't make sense to have a empty header in line 404.

hailiangzhang: Ok, a new commit pushed to handle empty user-agent headers; meanwhile, I found I didn't merge the existing user-agent headers, so provide the fix in the same commit.

tests/test_client.py
hailiangzhang: Alright, it's actually quite easy. I just added test case for POST and commit the change.

harmony/harmony.py
```diff
@@ -361,6 +362,57 @@ def _params(self, request: Request) -> dict:
...
+            session = self._session()
+            existing_user_agent_header = session.headers.get('User-Agent')
+            if existing_user_agent_header:
+                user_agent_content = set([existing_user_agent_header])
+            else:
+                user_agent_content = set([])
+
+            # Get harmony package info
+            try:
+                import harmony
```
chris-durbin, line 381: I think in most of our Python code we've been keeping our imports at the top of the file rather than inline.

harmony/harmony.py
```diff
@@ -361,6 +362,57 @@ def _params(self, request: Request) -> dict:
...
+                user_agent_content.add(f'harmony-py/{harmony.__version__}')
+            except Exception as e:
+                print("Following exception was caught "
+                      "when building user-agent headers for harmony-py:")
+                print(e)
+
+            # Get platform info
+            try:
+                p_system = platform.system()
```
chris-durbin, line 390: Just FYI - that with the harmony and harmony-service-lib PRs I don't think we're planning on adding OS type information to the user agent. We'll probably just have something like `harmony/0.0.0 harmony-uat`.

tests/test_client.py
```diff
@@ -263,6 +264,68 @@ def test_with_invalid_request():
...
+    assert responses.calls[0].request is not None
+    assert responses.calls[0].request.headers is not None
+    headers = responses.calls[0].request.headers
+    assert "User-Agent" in headers
+    user_agent_header = headers["User-Agent"]
+    # Since it's kinda overkill to find the exact character set
+    #   allowed in platform/implementation/version/etc,
+    #   the following regex may be a little bit more tolerant
+    assert re.match(
+        r"\s*([^/\s]+/[^/\s]+)(\s+[^/\s]+/[^/\s]+)*\s*", user_agent_header
```
chris-durbin, line 292: Since you use this regex in a couple of tests it's probably best to define a constant so if we decide to change it in the future we just change it in one place.

chris-durbin (approved):

harmony/harmony.py
hailiangzhang: Thanks for you comment Chris. It's probably a little bit strange to import harmony inside harmony, and I imported harmony here simply to get the version info, so I just put it inside this try block.  I can move it to the top if it doesn't seem to be strange or won't break things, or ideally, if there is a better way to get the harmony version info that will be great.

harmony/harmony.py
hailiangzhang: Thanks again Chris. Actually James pointed me the comment in JIRA story 767 with platform info included (https://bugs.earthdata.nasa.gov/browse/HARMONY-767), but please let me if you want me remove it.

tests/test_client.py
hailiangzhang: Good point ... will do.

harmony/harmony.py
chris-durbin: I don't really know Python and have no strong opinions. Maybe someone more knowledgable can weigh in.

harmony/harmony.py
chris-durbin: I think it makes sense you are providing that info especially for a client library. Maybe we should do the same thing for the harmony application too with the other ticket.

indiejames (approved):

harmony/harmony.py
hailiangzhang: Did a cleaner import ("`from harmony import __version__ as harmony_version`"), but still a little bit reluctant to put on the top of file (worries breaking other things if we can't import).

tests/test_client.py
hailiangzhang: Generated user-agent header regex inside a function instead of using the same literals in multiple places.

harmony/harmony.py
bilts: I don't understand why this is in a try block in the first place or why the `from` statement isn't at the top of the file, since the import should be completely under our control.  In what case could an exception be thrown?

bilts (approved):

harmony/harmony.py
hailiangzhang: Just moved this import to the top of the file.

PR #31, opened 2021-05-05
HARMONY-833: Actions for drafting and publishing releases

bilts:
Copies the equivalent code from harmony-service-lib-py

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,0 +1,38 @@
+name: Publish Release
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: '0'
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
+    - shell: bash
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: '0'
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
```
chris-durbin, line 16: I noticed the other repo is testing with 3.9. Should we be testing using the same versions and potentially multiple python versions like 3.6, 3.7, 3.8, and 3.9?

bilts: We are testing with those versions.  This is the python version used for building and publishing the artifact.  3.8 works.  I switched the other repo to use the same Python for consistency, though eventually I think it needs to have tests that mock access and can run in all versions.  I just wasn't ready overhaul the test suite to get releases set up.

vinnyinverso (approved):

chris-durbin (approved):

PR #41, opened 2021-12-21
Add human readable dates to status output

frankinspace:
Just a little quality of life update.

Added local datetime representation to status output to make it easier to read the times when viewed in a Jupyter notebook. Also updated tests to load examples from directory relative to the test_client.py file instead of relative to the current working directory where tests are being run.

Example of what the output of `status` returns after this change:

```
{
'created_at': datetime.datetime(2021, 2, 19, 18, 47, 31, 291000, tzinfo=tzutc()),
 'created_at_local': '2021-02-19T10:47:31-08:00',
 'message': 'The job is being processed',
 'num_input_granules': 32,
 'progress': 0,
 'request': 'https://harmony.earthdata.nasa.gov/{collection_id}/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?forceAsync=True&subset=lat(52%3A77)&subset=lon(-165%3A-140)&subset=time(%222010-01-01T00%3A00%3A00%22%3A%222020-12-30T00%3A00%3A00%22)',
 'status': 'running',
 'updated_at': datetime.datetime(2021, 2, 19, 18, 47, 31, 291000, tzinfo=tzutc()),
 'updated_at_local': '2021-02-19T10:47:31-08:00'
}
```

### DIFF
--- a/harmony/harmony.py
+++ b/harmony/harmony.py
@@ -588,12 +588,16 @@ class Client:
                 'numInputGranules'
             ]
             status_subset = {k: v for k, v in response.json().items() if k in fields}
+            created_at_dt = dateutil.parser.parse(status_subset['createdAt'])
+            updated_at_dt = dateutil.parser.parse(status_subset['updatedAt'])
             return {
                 'status': status_subset['status'],
                 'message': status_subset['message'],
                 'progress': status_subset['progress'],
-                'created_at': dateutil.parser.parse(status_subset['createdAt']),
-                'updated_at': dateutil.parser.parse(status_subset['updatedAt']),
+                'created_at': created_at_dt,
+                'updated_at': updated_at_dt,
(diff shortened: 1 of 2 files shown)


chris-durbin (commented):
Note this means the job status returned from `harmony-py` will have additional columns that aren't in the harmony response. I guess the question is do we need two different representations of the time (are they serving two different purposes for different types of users), or should we change the existing columns to be more useful for `harmony-py` users?

bilts:
> Note this means the job status returned from `harmony-py` will have additional columns that aren't in the harmony response. I guess the question is do we need two different representations of the time (are they serving two different purposes for different types of users), or should we change the existing columns to be more useful for `harmony-py` users?

I think this is something that can really only be done library-side, since we don't necessarily have access to the user's timezone on the server.  I see this more as a display helper.

PR #49, opened 2022-07-13
HARMONY-1125 - Allow dimensions to specify dim.min=0 or dim.max=0.

owenlittlejohns:
## Jira Issue ID

HARMONY-1125

## Description

Users may specify a dimension range with a minimum or maximum value that is 0. Prior to this change, when the request URL was constructed, there was a check for whether the minimum and maximum were falsy. If so, the value was assumed to mean `'*'`. However, 0 will also evaluate to `False`, meaning `Dimension('dim', 0, 10)` creates a subset parameter of `subset=dim(*:10)`, instead of `subset=dim(0:10)`. This PR addresses this edge-case.

## Local Test Steps

With the updated version of `harmony-py` create a request such as:

```python
from harmony import Client, Collection, Dimension, Environment, Request

harmony_client = Client(env=Environment.UAT)

hoss_named_dims_request = Request(collection=Collection(id='C1238392622-EEDTEST'),
                                  granule_id='G1245840464-EEDTEST',
                                  variables=['atmosphere_cloud_liquid_water_content'],
                                  dimensions=[Dimension('latitude', 0, 15),
                                              Dimension('longitude', -150, -105)])

harmony_client.submit(hoss_named_dims_request)
```
This should create a request URL that contains `subset=latitude(0:15)`, not `subset=latitude(*:15)`.

Note - this is mainly just to test named dimensions work with HOSS, I appreciate that `BBox` would normally be used in this example.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -48,10 +48,10 @@ def expected_full_submit_url(request):
     dimension_params = []
     if request.dimensions:
         for dim in request.dimensions:
-          name = dim.name
-          min = dim.min or '*'
-          max = dim.max or '*'
-          dimension_params += [f'subset={name}({min}:{max})']
+            name = dim.name
+            min = dim.min if dim.min is not None else '*'
+            max = dim.max if dim.max is not None else '*'
+            dimension_params += [f'subset={name}({min}:{max})']
 
     query_params = '&'.join(async_params + spatial_params + temporal_params + dimension_params)
     if request.format is not None:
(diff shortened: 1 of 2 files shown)


tests/test_client.py
```diff
@@ -48,10 +48,10 @@ def expected_full_submit_url(request):
     dimension_params = []
     if request.dimensions:
         for dim in request.dimensions:
-          name = dim.name
-          min = dim.min or '*'
-          max = dim.max or '*'
-          dimension_params += [f'subset={name}({min}:{max})']
+            name = dim.name
+            min = dim.min if dim.min is not None else '*'
+            max = dim.max if dim.max is not None else '*'
```
owenlittlejohns, line 53: I'm a bit wary of this test mechanism - it feels a bit odd to test some code by running identical code in the test module. (I'm likely missing something here, though)

chris-durbin (approved):

owenlittlejohns:
@chris-durbin - I was waiting for a second approval on this PR before merging. Should I continue to do so, or do you think this PR can be merged?

PR #50, opened 2022-09-09
Harmony 1224 partial results download

indiejames:
## Jira Issue ID
HARMONY-1224

## Description
This PR adds an `iterator()` method to the Client class that can be used to retrieve granules while a job is still processing.

## Local Test Steps
Check out this branch and run the examples/job_results_iterator.ipynb notebook.

## PR Acceptance Checklist
* [x] Acceptance criteria met
* [x] Documentation updated (if needed) - this happens automatically with the generated docs


harmony/harmony.py
```diff
@@ -968,6 +985,124 @@ def download_all(self,
...
+        Returns:
+            An Iterator that can be used to iterate over the granule results from a job
+        """
+        GET_JSON_RETRY_LIMIT = int(os.getenv('GET_JSON_RETRY_LIMIT', 3))
+        GET_JSON_RETRY_SLEEP = float(os.getenv('GET_JSON_RETRY_SLEEP', 1.0))
+        next_url = self._status_url(job_id)
+
+        # index used to keep track of where we are in the data links if the page gets reloaded
+        # so we don't pull the same granule more than once
+        # NOTE: this assumes that granule order is preserved when the page is reloaded and
```
chris-durbin, line 1040: When what page is reloaded? Do you mean that it is relying on the order of the links returned by harmony job status pages is consistent? I think that assumption is fine, but if that's what we're relying on maybe change the comment to:
```
 # index used to keep track of where we are in the data links if the page gets reloaded
 # so we don't pull the same data links more than once
 # NOTE: this relies on the link order being preserved when the harmony job status page
 # is reloaded and new links are added.
```

harmony/harmony.py
```diff
@@ -968,6 +985,124 @@ def download_all(self,
...
+                            'bbox': BBox(bbox[0], bbox[1], bbox[2], bbox[3]),
+                            'temporal': temporal
+                        }
+                elif link['rel'] == 'next':
+                    next_url = link['href']
+
+            if not next_url:
+                # no 'next' link means either we are done or we need to reload this page to
+                # get more results
+                if status in {'successful', 'paused', 'canceled', 'complete_with_errors'}:
```
chris-durbin, line 1087: Shouldn't we also stop on 'failed' status? I see a test for failed in the tests, so I'm not sure how that works - does 'failed' status cause some exception to happen and exit this code? Might be worth a comment on why 'failed' is not in this list.

tests/test_client.py
```diff
@@ -849,6 +851,210 @@ def test_download_all(mocker):
...
+
+def test_iterator_failed_job(mocker):
+    failed_job = expected_job('C123', 'abc123')
+    failed_job['status'] = 'failed'
+    failed_job['message'] = 'Job failed'
+    get_json_mock = mocker.Mock(return_value = failed_job)
+    mocker.patch('harmony.harmony.Client._get_json', get_json_mock)
+    client = Client(should_validate_auth=False)
+
+    # first iteration in which job state is 'running' and two granules have completed
```
chris-durbin, line 1036: I'm not following where the job starts out 'running' and two granules have completed. I think we should set the test up that way where a couple iterations work and then a third iteration the job is marked as failed and the iterator stops with an appropriate failure message, but I'm not sure where that's happening.

chris-durbin (commented):
When I tested the notebook all of the files downloaded at the same time for the first example job.

I believe that's because they were all started downloading in parallel and finished within the same 3 second interval? If that's not the case and the printing doesn't happen until the iterator finishes we need to fix that. I'll try another test with more granules that takes longer to make sure the behavior looks good.

examples/job_results_iterator.ipynb
```diff
@@ -0,0 +1,140 @@
...
+    "# the first argument to `iterator()` is the ID of the job we want\n",
+    "# the second argument is the path where we want to download the files\n",
+    "# the third argument is whether or not we want to overwrite files - this defaults to False which\n",
+    "# speeds up processing by not re-downloading files that have already been downloaded\n",
+    "iter = harmony_client.iterator(job_id, '', True)\n",
+    "# grab all the futures that are downloading the files\n",
+    "futures = list(map(lambda x: x['path'], iter))\n",
+    "# wait for all the futures to complete\n",
+    "(done_futures, _) = concurrent.futures.wait(futures)\n",
+    "#iterate of the futures to get the file paths\n",
```
chris-durbin, line 69: I'm not sure what this commenting is intending to say - probably can just be deleted.

indiejames: done

tests/test_client.py
indiejames: That comment was cut and paste from above and does not belong here - removed.

harmony/harmony.py
indiejames: We raise an exception when the job fails. Adding a comment as you requested.

harmony/harmony.py
indiejames: Done

indiejames:
After talking with Chris I added an optional print in the file download code so you can see the file paths as they get downloaded without waiting for everything to complete. This does not change how things are being downloaded and is mainly so you can easily see things happening in a notebook.

tests/test_client.py
chris-durbin: Can you make the test the way described - two successes and then third one the job status is failed and it exits out as expected? I just want to make sure we have a test that the iterator handles a change to the failed state when it didn't start out as failed.

vinnyinverso (approved):

chris-durbin (approved):

tests/test_client.py
indiejames: Done.

PR #68, opened 2023-06-27
HARMONY-1473: Added support for harmony capabilities endpoint.

ygliuvt:
## Jira Issue ID
HARMONY-1473

## Description
Added support for harmony capabilities endpoint in harmony-py.

## Local Test Steps
Run `collection_capabilities` notebook and verify it runs successfully.

## PR Acceptance Checklist
* [x] Acceptance criteria met


examples/collection_capabilities.ipynb
```diff
@@ -0,0 +1,159 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "harmony_client = Client(env=Environment.UAT)\n",
+    "\n",
+    "collection = Collection(id='C1234088182-EEDTEST')\n",
```
chris-durbin, line 47: We should add another cell that shows you can also specify the collection to use by providing a collection short name instead of a CMR concept ID.

harmony/harmony.py
```diff
@@ -277,6 +284,8 @@ def __init__(self,
         self.skip_preview = skip_preview
         self.ignore_errors = ignore_errors
         self.grid = grid
+        self.capabilities = capabilities
```
chris-durbin, line 287: Request is a class for submitting transformation requests using harmony. I don't think we should try to also use it to submit calls to get the collection capabilities. Probably best to add a new simple class specifically for getting capabilities that just has two fields - a Collection object and an optional capabilities version.

harmony/harmony.py
```diff
@@ -455,13 +465,16 @@ def _session(self):
...
     def _submit_url(self, request: Request) -> str:
         """Constructs the URL for the request that is used to submit a new Harmony Job."""
-        variables = [v.replace('/', '%2F') for v in request.variables]
-        vars = ','.join(variables)
-        return (
-            f'{self.config.root_url}'
-            f'/{request.collection.id}'
-            f'/ogc-api-coverages/1.0.0/collections/{vars}/coverage/rangeset'
-        )
+        if request.capabilities:
```
chris-durbin, line 468: Similar comment instead of calling `submit` on the harmony client - maybe add `get_capabilities` and pass in the Capabilities object as a parameter.

tests/test_client.py
```diff
@@ -1374,3 +1411,53 @@ def test_request_as_curl_post(examples_dir):
     assert f'https://harmony.earthdata.nasa.gov/{collection.id}' \
            f'/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset' in curl_command
     assert '-X POST' in curl_command
+
+@responses.activate
+def test_collection_capabilities():
```
chris-durbin, line 1416: Add one more test using collection short name instead of concept ID.

chris-durbin (commented):
Tested out the notebook successfully - just a couple refactor requests and additional demo/testing with using collection short name.

harmony/harmony.py
ygliuvt: Sure. I will refactor the code to add a CapabilitiesRequest.

harmony/harmony.py
ygliuvt: I think it would be an overkill once we make a different request type for capabilities request. The harmony client would still call `submit` to submit a CapabilitiesRequest and that will return the collection capabilities.

chris-durbin: The name `submit` might be confusing, but its intention (including all of the documentation) is to submit a transformation request using the OGC coverages API. It's not a generic `submit` as in call any arbitrary HTTP endpoint with a harmony root. I don't think we should overload it and then add if statements to change what API it calls.

examples/collection_capabilities.ipynb
ygliuvt: I am not even aware that we support ShortName as a parameter on capabilities endpoint. This is why we should have a more detailed API doc in Harmony. Filed ticket: harmony-1525.

harmony/harmony.py
indiejames: I'm okay with using `submit` if you are creating an entirely separate `CapabiliteisRequest` class that also has a `submit` method. Not sure of the semantics of what it would return though; since those requests are always synchronous it doesn't make sense to query the progress, etc.

ygliuvt: The main reason I want to reuse the `submit` function to handle two different requests is to reuse the common functions that are needed to for authentication and parameter handling. Since most of the request submission is the same for both requests, it make sense to share the functions rather than duplicate the functions. Using if statements is the easiest way to achieve this as Python lack the ability to overload functions by parameter type.

harmony/harmony.py
```diff
@@ -375,6 +394,52 @@ def error_messages(self) -> List[str]:
         return spatial_msgs + temporal_msgs + shape_msgs + dimension_msgs + parameter_msgs
 
 
+class CapabilitiesRequest(BaseRequest):
+    """A Harmony request to get the harmony capabilities of a CMR collection
+    Args:
+        A dict with optional collectionId, shortName and capabilities_version fields
```
chris-durbin, line 400: collection_id and short_name should be snake case

chris-durbin (approved):

harmony/harmony.py
```diff
@@ -173,7 +173,36 @@ def __repr__(self) -> str:
...
+
+    Args:
+        collection: The CMR collection that should be queried
+
+    Returns:
+        A Harmony Request instance
+    """
+
+    def __init__(self,
+                 collection: Collection):
```
indiejames, line 187: Given what we want for the hackfest, it probably doesn't make sense for `collection` to be a property of the base class, or at least not a required parameter of the constructor.

harmony/harmony.py
indiejames, line 400: I think it would be more idiomatic Python to just use keyword arguments here instead of taking a dictionary as a parameter

harmony/harmony.py
```diff
@@ -375,6 +394,52 @@ def error_messages(self) -> List[str]:
...
+
+    Returns:
+        A Harmony Capability Request instance
+    """
+
+    def __init__(self,
+                 request_params: dict
+                 ):
+
+        coll_identifier = request_params.get('collection_id')
```
indiejames, line 413: I think this would be a bit cleaner
```
coll_identifier = request_params.get'collection_id') or request_params.get('short_name')
```

indiejames: Or using the default value syntax of `get`
```
coll_identifier = request_params.get('collection_id', request_params.get('short_name'))
```

indiejames (approved):
Some stylistic comments that you can take or leave

harmony/harmony.py
ygliuvt: Makes sense. I will make it an optional keyword parameter.

PR #69, opened 2023-08-03
HARMONY-1414: Added support to harmony requests that completes successfully without creating a job.

ygliuvt:
## Jira Issue ID
HARMONY-1414

## Description
Added support to harmony requests that completes successfully without creating a job.

## Local Test Steps
Run last demo of tutorial notebook and verify the result returned from harmony client submission call is the download links in the harmony request response.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/harmony/harmony.py
+++ b/harmony/harmony.py
@@ -34,7 +34,7 @@ from contextlib import contextmanager
 from datetime import date, datetime
 from enum import Enum
 from typing import Any, ContextManager, IO, Iterator, List, Mapping, NamedTuple, Optional, \
-    Tuple, Generator
+    Tuple, Generator, Union
 
 import curlify
 import dateutil.parser
@@ -789,7 +789,9 @@ class Client:
             request: The Request to submit to Harmony (will be validated before sending)
 
         Returns:
-            The Harmony Job ID or download links or capability response
+            The Harmony Job ID for request done through async jobs
+            The JSON response for direct download request
+            The capabilities response for capabilities request
         """
         if not request.is_valid():
             msgs = ', '.join(request.error_messages())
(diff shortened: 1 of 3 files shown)


harmony/harmony.py
```diff
@@ -802,8 +802,10 @@ def submit(self, request: BaseRequest) -> any:
         if response.ok:
             if isinstance(request, CapabilitiesRequest):
                 return response.json()
+            elif response.json()['status'] == 'successful':
+                return response.json()['links']
```
chris-durbin, line 806: We should return the full JSON response for the no op service. When there's a job the user can use the job ID to pull back the full job status information, but in this case there's no way to get the other information with another call. I'd want to see the message for example that the download links do not reflect the subsetting I requested.

chris-durbin (changes requested):
I want to make sure a user can use the response similar to if a job were returned. In order to do that, I think we should update the response to include the full JSON rather than just the links. We should also update the download functions (including download_all) to take that JSON response and download the links. Ideally we can also make the wait_for_processing (or similarly named) function also work with that workflow.

ygliuvt:
> I want to make sure a user can use the response similar to if a job were returned. In order to do that, I think we should update the response to include the full JSON rather than just the links. We should also update the download functions (including download_all) to take that JSON response and download the links. Ideally we can also make the wait_for_processing (or similarly named) function also work with that workflow.

These are valid points and also completely changed the scope of the ticket. I will see how the download function works for directly download, update the ticket and implementation. Thanks.

indiejames (approved):

chris-durbin (approved):
Tested out successfully.

PR #74, opened 2023-10-30
Update Snyk vulnerabilities in documentation

owenlittlejohns:
## Jira Issue ID

N/A

## Description

I was in the Snyk UI and spotted `harmony-py` has a critical vulnerability in `requirements/example.txt`. This PR addresses the vulnerability by updating the dependencies for the notebooks in the `examples` directory. I also bumped the requirements in `requirements/docs.txt`, as there was a medium severity vulnerability there (`tornado` via `nbconvert`).

I checked things still worked by running the notebooks in the `examples` directory. Most ran, but there were a couple of exceptions:

* `job_stac.ipynb` couldn't find `pystac.STAC_IO`, so I changed that to `pystac.stac_io`, which seemed to work.
* The `helper.show_results` in `basics.ipynb` failed for a bunch of requests that were getting NetCDF-4 files out of HGA. I fixed this by forcing all the requests to give GeoTIFF outputs.
* There were some requests made to the Swath Projector that failed message validation. I thought Leo wrote a ticket for this, but can't find it in the DAS backlog. Anyway - that's not an issue with dependencies, but with the service itself.
* I was lazy and didn't try the in-region parts of the notebooks.

For the Sphinx documentation I had to tweak a method documentation string to make it all run again.

## Local Test Steps

For `requirements/examples.txt`:

* Pull this branch.
* Create a new Python environment (via conda or pyenv) and pip install `requirements/example.txt`.
* Start a Jupyter notebook server (`jupyter lab`).
* Run through the notebooks in the `examples` directory. Checking `basics.ipynb` is probably a good one, as it uses a lot of the functionality. Similiarly, the `job_stac.ipynb` is a good test of `pystac` and `intake-stac`.

For `requirements/docs.txt`:

* Pull this branch.
* Create a new Python environment (via conda or pyenv) and pip install `requirements/docs.txt`.
* Navigate to the `docs` directory.
* Run `make html` - this should complete successfully.
* Open the `docs/_build/html/index.html` file in a browser. It should look like some shiny Sphinx-generated documentation.

## PR Acceptance Checklist
* ~~Acceptance criteria met~~ N/A
* ~~Tests added/updated (if needed) and passing~~ N/A

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, '3.10', '3.11']
+        python-version: [3.8, 3.9, '3.10', '3.11']
 
     steps:
     - uses: actions/checkout@v3
--- a/requirements/core.txt
+++ b/requirements/core.txt
@@ -1,6 +1,6 @@
 python-dateutil ~= 2.8.2
 python-dotenv ~= 0.20.0
-progressbar2 ~= 3.55.0
+progressbar2 ~= 4.2.0
 requests ~= 2.28
 sphinxcontrib-napoleon ~= 0.7
 curlify ~= 2.2.1
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,7 +1,7 @@
-nbconvert ~= 7.4.0
+curlify ~= 2.2.1
 Jinja2 ~= 3.1.2
 load-dotenv ~=0.1.0
-curlify ~= 2.2.1
-progressbar2 ~= 3.55.0
-sphinx ~= 5.3.0
-sphinx-rtd-theme ~= 1.2.0
+nbconvert ~= 7.10.0
+progressbar2 ~= 4.2.0
+sphinx ~= 7.2.6
+sphinx-rtd-theme ~= 1.3.0
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -3,5 +3,5 @@ Jinja2 ~= 3.1.2
 load-dotenv ~=0.1.0
 nbconvert ~= 7.10.0
 progressbar2 ~= 4.2.0
-sphinx ~= 7.2.6
+sphinx ~= 7.1.2
 sphinx-rtd-theme ~= 1.3.0
--- a/requirements/examples.txt
+++ b/requirements/examples.txt
@@ -1,10 +1,10 @@
-boto3 ~= 1.17
-ipyplot ~= 1.1
-ipywidgets ~= 7.6
-jupyterlab ~= 3.0
-matplotlib ~= 3.3
-pystac ~= 1.7.3
-rasterio ~= 1.2
+boto3 ~= 1.28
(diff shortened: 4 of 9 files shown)


owenlittlejohns:
Quick update - I was getting failing builds because I updated the `progressbar2` dependency in `requirements/docs.txt`, but another version was present in `requirements/core.txt`. I've updated both places, and verified the tests still pass (locally at least)

.github/workflows/tests.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, '3.10', '3.11']
```
owenlittlejohns, line 10: The update to a recent version of `nbconvert` requires Python >= 3.8. I also checked and 3.7 was EOL back in June (see: https://devguide.python.org/versions/). It would have been great to not drop 3.7 in the testing, but this seems to be the answer to the rabbit hole of dependencies. (The root of the issue comes from the `tests.yml` workflow installing the core, test and docs dependencies, and therefore needing all three to be consistent with one another)

requirements/core.txt
```diff
@@ -1,6 +1,6 @@
 python-dateutil ~= 2.8.2
 python-dotenv ~= 0.20.0
-progressbar2 ~= 3.55.0
+progressbar2 ~= 4.2.0
```
owenlittlejohns, line 3: I wanted to leave these dependencies entirely alone, but `progressbar2` needs to be consistent in `core.txt` and `docs.txt` to ensure the `tests.yml` workflow can run. The `sphinx` dependency in `dev.txt` was bumped for the same reason.

requirements/examples.txt
```diff
@@ -1,10 +1,11 @@
...
 intake-stac ~= 0.4.0
-netCDF4
-numpy
+ipyplot ~= 1.1
+ipywidgets ~= 8.1
+jupyterlab ~= 4.0
+matplotlib ~= 3.8
+netCDF4 ~= 1.6
+numpy ~= 1.26
+pillow ~= 10.1  # A dependency of ipyplot, pinned to avoid critical vulnerability.
```
owenlittlejohns, line 9: I'm guessing future releases of `ipyplot` will bump their own required version of `pillow`, so this line can be removed at a later date. 🤞

requirements/docs.txt
```diff
@@ -1,7 +1,7 @@
...
+curlify ~= 2.2.1
 Jinja2 ~= 3.1.2
 load-dotenv ~=0.1.0
-curlify ~= 2.2.1
-progressbar2 ~= 3.55.0
-sphinx ~= 5.3.0
-sphinx-rtd-theme ~= 1.2.0
+nbconvert ~= 7.10.0
+progressbar2 ~= 4.2.0
+sphinx ~= 7.1.2
```
owenlittlejohns, line 6: `sphinx` goes up to 7.2.6, but 7.2.0 onwards all require Python 3.9+. 7.1.2 was only released in August, and currently has no vulnerabilities.

ygliuvt (approved):
Verified the readthedocs looks good.

PR #81, opened 2024-03-13
HARMONY-1721: Switch from using GET to a POST in all OGC coverages calls to harmony

chris-durbin:
## Jira Issue ID
HARMONY-1721

## Description
harmony-py was using a GET for all calls to harmony, but this was causing 413 issues when trying to use a long list of granuleIds because of URL length limits. This PR switches to use a POST with multipart/form-data (we were already doing this when a shapefile was included in the request).

## Local Test Steps
Run `make test` and verify all tests pass.
Run the first request in the `examples/tutorial.ipynb` notebook. Note that there's a bug in harmony parsing width and height which will be addressed as a separate ticket. So you'll need to comment those out so the cell looks like this:
```
collection = Collection(id='C1234088182-EEDTEST')

request = Request(
    collection=collection,
    spatial=BBox(-165, 52, -140, 77),
    temporal={
        'start': dt.datetime(2010, 1, 1),
        'stop': dt.datetime(2020, 12, 30)
    },
    variables=['blue_var'],
    max_results=10,
    crs='EPSG:3995',
    format='image/tiff',
    # height=512,
    # width=512
)

request.is_valid()
```
Verify the request completes successfully. You can also check the harmony logs in the UAT environment to tell that it is using a POST and not a GET.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/harmony/harmony.py
+++ b/harmony/harmony.py
@@ -174,7 +174,8 @@ _valid_shapefile_exts = ', '.join((_shapefile_exts_to_mimes.keys()))
 
 
 class BaseRequest:
-    """A Harmony base request with the CMR collection. It is the base class of all harmony requests.
+    """A Harmony base request with the CMR collection. It is the base class of all harmony
+    requests.
 
     Args:
         collection: The CMR collection that should be queried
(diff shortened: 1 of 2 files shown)


ygliuvt (approved):
Tested successfully locally.

indiejames (approved):
Ran tests and tutorial notebook successfully

PR #85, opened 2024-06-06
HARMONY-1762: Switch to passing variables in the query parameters

chris-durbin:
## Jira Issue ID
HARMONY-1762

## Description
Change to pass the variables in the query parameters using `parameter_vars` so that we can support requests with 4000+ variables.

## Local Test Steps
I tested with the intro_tutorial notebook and modifying the request to use a bounding box instead of a shapefile (the request_as_url and request_as_curl methods do not work with a shapefile).

Add a couple print statements to the notebooks to see what the URL looks like:
print(harmony_client.request_as_url(request))
print(harmony_client.request_as_curl(request))

Make sure you see it uses `parameter_vars` and the variable is included in the query parameters.

Also test removing the variables altogether and make sure it passes `variable=all`.

It would be good to test with a large number of variables as well if someone has a good request for that.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -137,3 +137,4 @@ examples/*.tif
 __pypackages__/
 
 .idea
+.virtual_documents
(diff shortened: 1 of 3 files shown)


ygliuvt (approved):
Tested a couple notebooks and the harmony requests still works.

vinnyinverso (approved):

PR #102, opened 2025-02-05
HARMONY-1898: As harmony-py user, I want to be able add one or more labels to one or more of my jobs

ygliuvt:
## Jira Issue ID
HARMONY-1898

## Description
As harmony-py user, I want to be able add one or more labels to one or more of my jobs

## Local Test Steps
Run labels notebook and verify adding labels to jobs work as expected. 
Verify readthedocs documentation is ok.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/harmony/harmony.py
+++ b/harmony/harmony.py
@@ -237,8 +237,17 @@ _shapefile_exts_to_mimes = {
 _valid_shapefile_exts = ', '.join((_shapefile_exts_to_mimes.keys()))
 
 
+class HttpMethod(Enum):
+    """The http method of a harmony request.
+    """
+    GET = "GET"
+    PUT = "PUT"
+    POST = "POST"
+    DELETE = "DELETE"
+
+
 class BaseRequest:
-    """A Harmony base request with the CMR collection. It is the base class of all harmony
+    """A Harmony base request for all client requests. It is the base class of all harmony
     requests.
 
     Args:
@@ -250,9 +259,8 @@ class BaseRequest:
 
     def __init__(self,
                  *,
-                 collection: Collection):
-        self.collection = collection
-        self.variable_name_to_query_param = {}
+                 http_method: HttpMethod = HttpMethod.GET):
+        self.http_method = http_method
 
     def error_messages(self) -> List[str]:
         return []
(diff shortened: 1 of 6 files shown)


chris-durbin (approved):
Tested successfully, I like the refactor to BaseRequest. I also verified we have a separate ticket for removing labels from jobs.

harmony/harmony.py
```diff
@@ -552,17 +586,57 @@ def error_messages(self) -> List[str]:
...
+class LabelsRequest(BaseRequest):
+    """A Harmony request to create or delete labels for jobs.
+
+    This request allows users to associate labels with Harmony jobs or remove
+    existing labels, facilitating job management and categorization.
 
-    s3: Returns an Amazon Web Services (AWS) S3 URL
-    https: Returns a standard HTTP URL
+    Args:
+        http_method (HttpMethod): The HTTP method to use for the request (e.g., PUT, DELETE).
```
indiejames, line 596: I'm willing to be overruled, but I don't like the user specifying the `http_method` here. It is an implementation detail. Also, trying `http_method=HttpMethod.DELETE` gives an error. I would prefer you have separate request types, `AddLablesRequest`, `DeleteLabelsRequest`, but I would settle for using something like `action` instead of `http_method`, where `action` is an enum type defined elsewhere. For now the delete request or delete action would just not be implemented, so no one would be tempted to try deleting yet.

indiejames (commented):
See my comment about the `http_method` parameter.

harmony/harmony.py
ygliuvt: I thought about this when implementing. Definitely don't want the action in request name. It's a pain to introduce new Request types. I thought about an `action` attribute, but didn't go that way because it is too loose and indirect, mainly because I don't want to deal with `add`, `create` vs `update` and their meaning on other routes/request types. I can see your point with actually prefer `action` over `http_method`. I can change it to `action` if you feel strongly.

chris-durbin: Agreed - my first thought also was that `AddLabelsRequest` and `DeleteLabelsRequest` would be better than an `action` parameter.

ygliuvt: OK. Refactored code.

indiejames (approved):
Tested code changes - looks good

PR #108, opened 2025-03-04
HARMONY-1997: As a harmony-py user I would like to see a warning message when a job results in no data.

ygliuvt:
## Jira Issue ID
HARMONY-1997

## Description
Updated harmony-py to report error when there is no files to download for failed or warning requests during `client.download_all` invocation.

## Local Test Steps
Since we have to use customized (fake) service image to get requests into warning state, it doesn't make sense to add a notebook to demonstrate this behavior. Also the current failed request notebook is returning success probably due to recent configuration changes which we will have to fix as a separate issue (HARMONY-2034).

Here is a notebook snippet to test the error reporting for warning request. I will restore the faked service-example image in UAT once this PR is merged.  

```
import helper
helper.install_project_and_dependencies('..', libs=['examples'])
import json
import sys
import datetime as dt
from harmony import BBox, Client, Collection, Request, Environment

harmony_client = Client(env=Environment.UAT)

collection = Collection(id='C1233800302-EEDTEST')

request = Request(
    collection=collection,
    spatial=BBox(-140, 20, -50, 60),
    granule_id='G1233800343-EEDTEST',
    crs='EPSG:31975',
    variables=['blue_var'],
    format='image/png'
)

job_id = harmony_client.submit(request)

for filename in [f.result() for f in harmony_client.download_all(job_id)]:
    print(f'\nDownload file: {filename}')
```

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/harmony/client.py
+++ b/harmony/client.py
@@ -887,6 +887,24 @@ class Client:
                 print(filename)
             return filename
 
+    def job_status_message(self, job_id: str) -> str:
+        """Extract the job status and message from job results.
+
+        Blocks until the Harmony job is done processing.
+
+        Args:
+            job_id: UUID string for the job you wish to interrogate.
+
+        Returns:
+            the job status and message.
+
+        :raises
+            Exception: This can happen if an invalid job_id is provided or Harmony services
+            can't be reached.
+        """
+        data = self.result_json(job_id)
+        return data['status'], data['message']
+
     def download(self, url: str, directory: str = '', overwrite: bool = False) -> Future:
         """Downloads data and saves it to a file asynchronously.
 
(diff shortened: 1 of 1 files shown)


harmony/client.py
```diff
@@ -940,10 +958,23 @@ def download_all(self,
...
+                    if url.endswith('zarr'):
+                        raise self.zarr_download_exception
+                    yield self.executor.submit(self._download_file, url, directory, overwrite)
+            except ProcessingFailedException:
+                print('\nJob Status is failed. There is no file to download.', file=sys.stderr)
+                # set num_files to 1 to bypass the extra error reporting below
+                # as it has already been reported
+                num_files = 1
+
+            if num_files == 0:
```
chris-durbin, line 974: It looks like we could move this check into the try block and not do the workaround on line 972.

chris-durbin (approved):
Tested successfully.

harmony/client.py
ygliuvt: Refactored code.

indiejames (approved):

PR #109, opened 2025-03-12
HARMONY-2035: As a harmony-py user, I want to request pixel subsetting.

ygliuvt:
## Jira Issue ID
HARMONY-2035

## Description
This PR is the harmony-py part to support client requesting pixel subsetting. There is a corresponding harmony PR that should be used to test the harmony-py change.
It adds a new `pixel_subset` parameter in  harmony-py request to support pixel subsetting.

## Local Test Steps
The `pixel_subset` parameter can be unset, `True` or `False` in a harmony-py request. Feel free to experiment with different values in the following tests.

Use the following block to test OGC Rangeset requests:
```
import helper
helper.install_project_and_dependencies('..', libs=['examples'])
import json
import sys
import datetime as dt
from harmony import BBox, Client, Collection, Request, Environment

harmony_client = Client(env=Environment.LOCAL)
collection = Collection(id='C1233800302-EEDTEST')
request = Request(
    collection=collection,
    spatial=BBox(-140, 20, -50, 60),
    granule_id='G1233800343-EEDTEST',
    crs='EPSG:31975',
    variables=['blue_var'],
    format='image/png',
    pixel_subset=True
)

url = harmony_client.request_as_url(request)
print(url)

job_id = harmony_client.submit(request)

for filename in [f.result() for f in harmony_client.download_all(job_id)]:
    print(f'\nDownload file: {filename}')
```

Use the following block to test OGC EDR requests:
```
import helper
helper.install_project_and_dependencies('..', libs=['examples'])
import json
import sys
import datetime as dt
from harmony import WKT, BBox, Client, Collection, Request, Environment

harmony_client = Client(env=Environment.LOCAL)
collection = Collection(id='C1233800302-EEDTEST')

request = Request(
    collection=collection,
    spatial=WKT('POLYGON((-140 20, -50 20, -50 60, -140 60, -140 20))'),
    granule_id=['C1233800302-EEDTEST'],
    max_results=1,
    temporal={
        'start': dt.datetime(1980, 1, 1),
        'stop': dt.datetime(2020, 12, 30)
    },
    variables=['blue_var'],
    crs='EPSG:31975',
    format='image/png',
    pixel_subset=False
)

url = harmony_client.request_as_url(request)
print(url)

job_id = harmony_client.submit(request)
for filename in [f.result() for f in harmony_client.download_all(job_id)]:
    print(f'\nDownload file: {filename}')
```

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/harmony/client.py
+++ b/harmony/client.py
@@ -262,17 +262,6 @@ class Client:
 
         return params
 
-    def _params_to_json(self, params: Mapping[str, Any]) -> Mapping[str, Any]:
-        """Returns the given parameter mapping as a JSON POST body
-        Args:
-            params: A dictionary of parameter mappings as returned by self._params(request)
-
-        Returns:
-            A JSON map that can be submitted as a POST body
-        """
-        return {key: (value == 'true' if key in boolean_params else value)
-                for key, value in params.items()}
-
     def _headers(self) -> dict:
         """
         Create (if needed) and return a dictionary of headers.
@@ -427,7 +416,7 @@ class Client:
                 if request.is_edr_request():
                     r = requests.models.Request('POST',
                                                 self._submit_url(request),
-                                                json=self._params_to_json(params),
+                                                json=params,
                                                 headers=headers)
                 else:
                     param_items = self._params_dict_to_files(params)
--- a/harmony/client.py
+++ b/harmony/client.py
@@ -57,6 +57,8 @@ progressbar_widgets = [
     ' [', progressbar.RotatingMarker(), ']',
 ]
 
+boolean_params = ['forceAsync', 'concatenate', 'skipPreview', 'ignoreErrors', 'pixelSubset']
+
 
 def temporal_to_edr_datetime(temporal: dict) -> str:
     datetime_format = '%Y-%m-%dT%H:%M:%SZ'
@@ -222,7 +224,9 @@ class Client:
         skipped_params = ['shapefile']
         if isinstance(request, OgcBaseRequest):
             if request.is_edr_request():
-                params['forceAsync'] = True
+                # use string in lowercase as value to match how boolean values
+                # in params are converted below
+                params['forceAsync'] = 'true'
                 if request.spatial:
                     params['coords'] = request.spatial.wkt
                 if request.temporal:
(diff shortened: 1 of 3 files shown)


tests/test_request.py
```diff
@@ -369,3 +369,19 @@ def test_valid_get_jobs_request_with_multiple_labels_in_string():
...
+
+def test_request_with_pixel_subset_true():
+    request = Request(collection=Collection('foobar'), pixel_subset=True)
+    assert request.is_valid()
+    assert request.pixel_subset is not None and request.pixel_subset == True
+
+
+def test_request_defaults_to_pixel_subset_none():
+    request = Request(collection=Collection('foobar'))
+    assert request.pixel_subset is None
```
chris-durbin, line 387: I'd like to test that setting pixel_subset to a non-boolean value causes request.is_valid() to be False and an appropriate error message is provided.

harmony/client.py
```diff
@@ -222,7 +224,9 @@ def _params(self, request: BaseRequest) -> dict:
         skipped_params = ['shapefile']
         if isinstance(request, OgcBaseRequest):
             if request.is_edr_request():
-                params['forceAsync'] = True
+                # use string in lowercase as value to match how boolean values
+                # in params are converted below
+                params['forceAsync'] = 'true'
```
chris-durbin, line 229: I'd prefer if we used booleans throughout for all boolean parameters and treat it as an error if a string or any other type is used.

We can add the boolean validations to this function:
``` 
def error_messages(self) -> List[str]:
        """A list of error messages, if any, for the request."""
```

chris-durbin (commented):
I found setting pixel_subset='Foo' was allowed and sent the request to harmony. I'd like if we enforced that all booleans must be booleans and that we set an error message indicating that if someone tries to set it to anything other than True or False (possibly None but we'd have to make sure that works correctly and does not add the parameter to the request).

harmony/client.py
ygliuvt: This is used to submit to Harmony which has to be a string now.

tests/test_request.py
ygliuvt: Added boolean parameter validations.

harmony/request.py
```diff
@@ -420,7 +426,15 @@ def __init__(self,
         ]
         self.parameter_validations = [  # for simple, one-off validations
             (True if self.destination_url is None else self.destination_url.startswith('s3://'),
-             ('Destination URL must be an S3 location'))
+             'Destination URL must be an S3 location'),
+            (self.concatenate is None or isinstance(self.concatenate, bool),
+             'concatenate must be either True of False'),
```
chris-durbin, line 431: 'True of False' typo in several places.

Since we are requiring a boolean type (`pixel_subset='True'` fails validation), we should say something like:
'concatenate must be a boolean type either True or False'

chris-durbin (approved):
Tested successfully, just a comment on changing the validation error message.

harmony/request.py
ygliuvt: Updated message to `concatenate must be a boolean (True or False)`

indiejames (approved):

PR #110, opened 2025-03-26
HARMONY-2026: As a client, I want to cancel a job using harmony-py.

ygliuvt:
## Jira Issue ID
HARMONY-2026

## Description
Add support in harmony-py to cancel a job.
Add cancel job example in Jupyter notebook.
Fix broken job preview example in Jupyter notebook.

## Local Test Steps
run `job_pause_resume.ipynb` notebook

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/examples/job_pause_resume.ipynb
+++ b/examples/job_pause_resume.ipynb
@@ -112,7 +112,15 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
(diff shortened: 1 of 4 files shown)


chris-durbin (commented):
I'm having issues right now with harmony-py preventing me from running the notebooks. I did see that there are a couple of cells that generate exceptions which cause the notebook to stop executing when you try to run the entire thing at once. We should update both of those cells to use try blocks and catch the exceptions.

> #### Attempting to pause a completed job will result in an error

> #### Attempting to resume a completed job will also result in an error

indiejames (approved):

PR #119, opened 2025-10-06
DAS-2427: Adds retry logic to harmony requests Session

flamingbear:
## Jira Issue ID

[DAS-2427](https://bugs.earthdata.nasa.gov/browse/DAS-2427) 

## Description

NSIDC testers were getting 502 and 504 errors during some long lasting requests. This caused some confusion as to why they would get an error in their notebook, but then go and look at the workflow-ui and see that the harmony service succeeded (or was still READY or RUNNING). 

This PR adds retries to those transient errors.

Here are some questions since I had some gemini help becuase I'm not an expert.

1. Should all of the errors retried be retried? I have [429, 500, 502, 503, 504]
   - These were definitely suggested by LLM.
   - I see conflicting information about 429, too many requests (we don't want to retry too soon because that just adds more requests I would guess, but it would definitely handle the request eventually.)
   - I see that 502 shouldn't be retried, but I think we're getting it for temporary errors and I can find other suggestions that it should. 

2. Is backoff_factor 1 good?
   - It seems long so that 7 seconds would elapse before 3 Retries would occur.  I could make it shorter, or make it shorter and add a few more tries? I just don't really know.
   - I added backoff_jitter just because it says that's best practices. 

## Local Test Steps

Build the branch and run the tests.
```
make install
make lint
make test
```

Next you can build a distribution and install to an envronment. 

```
make build
```

That creates a .whl and .gz distribution. I installed it to my regression tests
and ran them against localhost and UAT envs. Just as a sanity check.

Let me know if there are other things to test/update?

## PR Acceptance Checklist
* [ ] ~Acceptance criteria met~
* [ ] ~Documentation updated (if needed)~

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1601,6 +1601,36 @@ def test_handle_error_response_invalid_json():
     # Check retries
     assert len(responses.calls) == 9 # (1 + 4 + 4)
 
+@responses.activate
+def test_handle_non_transient_error_no_retry():
+    job_id = '89733-badc-1324'
+    collection = Collection(id='F229040468263-STRP')
+    request = Request(
+        collection=collection,
+        spatial=BBox(-107, 40, -105, 42)
+    )
+    responses.add(
+        responses.POST,
+        expected_submit_url(collection.id),
+        status=200,
+        json=expected_job(collection.id, 'abcd-1234')
+    )
+    responses.add(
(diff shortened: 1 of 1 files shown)


flamingbear:
I added this as ready for review, but it's ready for discussion more than anything.

PR #125, opened 2026-02-16
Fix #123: Add validation for invalid temporal keys like 'end'

### LINKED ISSUE #123: `Request.is_valid()` passes with invalid temporal key "end", causing silent failure and backend errors
I encountered a confusing error when submitting a Harmony request where `harmony-py` silently ignored a portion of my temporal definition, leading to a job failure on the backend.

I accidentally defined my temporal range using the key `'end'` instead of `'stop'`:
```python
time_range = {
    'start': datetime(2019, 11, 1),
    'end': datetime(2019, 12, 30) 
}
```

**Expected Behavior:**
The `harmony-py` client should either:
1. Accept `'end'` as an alias for `'stop'` (since it is a common intuitive keyword and as far as I understand very similar to the actual CMR key `EndDate`; see [here](https://wiki.earthdata.nasa.gov/spaces/CMR/pages/115937354/Temporal+Extents)), OR
2. Raise a `ValidationError` or warning when `Request.is_valid()` is called, alerting me that the key `'end'` is unknown and will be ignored.

**Actual Behavior:**
1. `Request.is_valid()` returned `True`, erroneously confirming the request was correct.
2. `harmony-py` constructed a request URL with an open-ended time range (e.g., `subset=time("2019-11-01T...":*)`), effectively dropping the end date.
3. The job was submitted successfully. `results_json = harmony_client.result_json(job_id, show_progress=True)` then returns something like the following where it is just hinted at something being wrong "Job is running with errors":

```
 [ Processing:   9% ] |####                                               | [|]
 [ Processing:   9% ] |####                                               | [/]
 [ Processing:   9% ] |####                                               | [-]
 [ Processing:   9% ] |####                                               | [\]
 [ Processing:   9% ] |####                                               | [|]

Job is running with errors.
 [ Processing:  10% ] |#####                                              | [/]
 [ Processing:  10% ] |#####                                              | [-]
 [ Processing:  10% ] |#####                                              | [\]
 [ Processing:  10% ] |#####                                              | [|]
```
Finally, `harmony_client.download_all(job_id)` is then just downloading 0 files and that's it. Only when looking at `results_json` it becomes clear what the actual problem was... specifically `nasa/harmony-trajectory-subsetter` failed with the error: `Invalid temporal range, both start and end required` for each file that was supposed to be processed and downloaded.

So this ended up being a confusing debugging experience that might be similar for other parameters / data products and should be evaluated in my opinion.

sonikaarora:
## Jira Issue ID
GitHub Issue #123

## Description

`Request.is_valid()` was incorrectly passing when users accidentally used the 
key `'end'` instead of `'stop'` in temporal range dictionaries. This caused 
silent failures where:
- Validation passed erroneously
- The 'end' key was silently ignored
- Backend services received incomplete temporal ranges
- Jobs failed with confusing errors like "Invalid temporal range, both start 
  and end required"

### Solution
Added validation to the `Request` class temporal validations that:
- Checks that temporal dictionary keys are only 'start' or 'stop'
- Provides a clear, actionable error message when invalid keys are detected
- Specifically mentions the common mistake of using 'end' instead of 'stop'

### Changes
- **harmony/request.py**: Added validation check in `temporal_validations` list
- **tests/test_request.py**: Added 3 comprehensive test cases covering:
  - Using 'start' and 'end' together
  - Using only 'end' without 'start'
  - Using multiple invalid keys including 'end
  - 
## PR Acceptance Checklist
* [ ] Acceptance criteria met

### DIFF
--- a/harmony/request.py
+++ b/harmony/request.py
@@ -424,8 +424,7 @@ class Request(OgcBaseRequest):
              ('When included in the request, the temporal range should include a '
               'start or stop attribute.')),
             (lambda tr: all(key in ['start', 'stop'] for key in tr.keys()),
-             ('Temporal range keys must be either "start" or "stop". '
-              'Note: Use "stop" instead of "end" (common mistake).')),
+             ('Temporal range keys must be either "start" or "stop".')),
             (lambda tr: tr['start'] < tr['stop'] if 'start' in tr and 'stop' in tr else True,
              'The temporal range\'s start must be earlier than its stop datetime.')
         ]
--- a/tests/test_request.py
+++ b/tests/test_request.py
@@ -211,22 +211,19 @@ def test_request_spatial_error_messages(key, value, message):
             'start': dt.datetime(2019, 11, 1),
             'end': dt.datetime(2019, 12, 30)
         },
-        ('Temporal range keys must be either "start" or "stop". '
-         'Note: Use "stop" instead of "end" (common mistake).')
+        ('Temporal range keys must be either "start" or "stop".')
     ), (
         'temporal', {
             'end': dt.datetime(2019, 12, 30)
         },
-        ('Temporal range keys must be either "start" or "stop". '
-         'Note: Use "stop" instead of "end" (common mistake).')
+        ('Temporal range keys must be either "start" or "stop".')
     ), (
         'temporal', {
             'start': dt.datetime(2019, 11, 1),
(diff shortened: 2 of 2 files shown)


chris-durbin (commented):
Thanks for the PR - just one comment about simplifying the error message, otherwise looks good.

PR #132, opened 2026-06-03
fix: remove upper bounds from runtime dependencies

### LINKED ISSUE #79: Consider removing upper bounds from dependencies in setup.py

stark256-spec:
## Summary

Fixes #79

### Problem

All runtime dependencies in `pyproject.toml` used `~=X.Y.Z` (compatible release) pins, which Python resolves as `>=X.Y.Z, <X.Y+1`. For a library like `harmony-py` that is installed into user environments alongside other packages, this creates unnecessary dependency conflicts whenever any transitive dependency requires a newer minor version of the same package.

### Fix

Replace all `~=` specifiers in `[project.dependencies]` with `>=` lower bounds, retaining only the minimum version known to work. The `shapely <3` upper bound is also removed as no incompatibility with shapely 3.x has been identified.

Dev and docs optional dependencies are left unchanged as those are developer-controlled environments where tighter pins are acceptable.

## Test plan

- [ ] Existing test suite passes with relaxed bounds
- [ ] No new import or runtime errors observed

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,11 @@
 name: Tests
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    # Nightly run at 02:00 UTC — catches breakage from new upstream releases
+    - cron: '0 2 * * *'
 
 jobs:
   build:
--- a/harmony/request.py
+++ b/harmony/request.py
@@ -220,6 +220,9 @@ def is_wkt_valid(wkt_string: str) -> bool:
         # Handle WKT reading errors and invalid WKT strings
         print(f"Invalid WKT: {e}")
         return False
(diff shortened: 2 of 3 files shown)


flamingbear:
I like this, but I'd like to make sure we're not going to run into any issues if we try to run with the latest versions available and keep up with that on a nightly basis.  Can you make sure the build and tests actions run every night on this repo?

stark256-spec:
Added a nightly cron schedule (`0 2 * * *` — 02:00 UTC) to `.github/workflows/tests.yml`. The full Python matrix (3.9 → 3.13) will now run every night against whatever versions pip resolves, so any breakage from a new upstream release surfaces the same day.

flamingbear:
This will keep your contribution but pull it into #133

flamingbear:
closing for parallel PR #133

PR #133, opened 2026-06-03
HARMONY-1837 Relax dependency upper bounds

### LINKED ISSUE #79: Consider removing upper bounds from dependencies in setup.py

flamingbear:
## Description 

This PR pulls in the user contribution changes for #132 

This will close #79 and HARMONY-1837

Code Changes: 
 - Unpins top bounds on dependencies.
 - Adds new requirements-min.txt to test minimum requirements
 - Updates the workflow to test both latest and minimum requirements
 - adds an install-minimum to the Makefile to constrain the package install to the pinned values.

Workflow changes: Adds a notify_nightly_failure task that will post a message to a slack webhook. Right now that webhook is configured to an NSIDC channel that I will listen to, but I have requested an app be created in the NASA org that we can use to post to #dev-harmony-only

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,7 +42,8 @@ jobs:
     runs-on: ubuntu-latest
     if: >
       always() &&
-      github.event_name == 'schedule' &&
+      (github.event_name == 'schedule' ||
+       github.event_name == 'workflow_dispatch') &&
       needs.build.result == 'failure'
 
     steps:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,6 +3,7 @@ name: Tests
 on:
   push:
   pull_request:
+  workflow_dispatch: # for testing remove
   schedule:
     # Nightly run at 02:00 UTC — catches breakage from new upstream releases
     - cron: '0 2 * * *'
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,6 +17,10 @@ jobs:
       matrix:
         python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
         dependencies: ['minimum', 'latest']
+      exclude:
+        # shapely 2.0.4 has no pre-build wheel for 3.13
+        - python-version: '3.13'
+          dependencies: 'minimum'
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,8 +7,7 @@ on:
   pull_request:
   workflow_dispatch:
   schedule:
-    # Nightly run at 02:00 UTC
-    - cron: '0 2 * * *'
+    - cron: '0 2 * * 1,3,5'
 
 jobs:
   build:
@@ -40,7 +39,7 @@ jobs:
         path: htmlcov/*
       if: ${{ github.event_name == 'pull_request' && (github.event.action == 'opened' || github.event.action == 'synchronize') }}
 
-  notify_failure:
+  notify_nightly_failure:
     needs: build
     runs-on: ubuntu-latest
     if: >
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,11 @@
 name: Tests
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    # Nightly run at 02:00 UTC — catches breakage from new upstream releases
+    - cron: '0 2 * * *'
 
 jobs:
   build:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ on:
   pull_request:
   workflow_dispatch:
   schedule:
-    # Nightly run at 02:00 UTC — catches breakage from new upstream releases
+    # Nightly run at 02:00 UTC
     - cron: '0 2 * * *'
 
 jobs:
@@ -40,7 +40,7 @@ jobs:
         path: htmlcov/*
       if: ${{ github.event_name == 'pull_request' && (github.event.action == 'opened' || github.event.action == 'synchronize') }}
 
-  notify:
+  notify_failure:
     needs: build
     runs-on: ubuntu-latest
     if: >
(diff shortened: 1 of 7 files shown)


flamingbear:
I don't love the inline script for the notification. I'm going to put this back into draft and wait 'til I can focus on it a bit. maybe IP.

harmony/request.py
```diff
@@ -220,6 +220,9 @@ def is_wkt_valid(wkt_string: str) -> bool:
         # Handle WKT reading errors and invalid WKT strings
         print(f"Invalid WKT: {e}")
         return False
+    except (ShapelyError, NotImplementedError) as e:
```
chris-durbin: This was duplicated

flamingbear: I don't think so, one is catching `ValueError` and this is catching a new `NotImplementedError` that was added to [shapely in 2.0.7](https://github.com/shapely/shapely/releases/tag/2.0.7). (I thought I added a note somewhere else)

.github/workflows/tests.yml
```diff
@@ -1,13 +1,26 @@
...
   build:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        dependencies: ['minimum', 'latest']
+        exclude:
+          # shapely 2.0.4 has no pre-build wheel for 3.13
+          - python-version: '3.13'
+            dependencies: 'minimum'
```
flamingbear, line 23: This will update when we drop support for 3.9 in #138

tests/test_client.py
```diff
@@ -1396,6 +1396,8 @@ def side_effect_for_get_json_failed_job(extra_links) -> List[str]:
 def test_iterator_failed_job(link_type, mocker):
     # test with two successful work items followed by a failed job
     extra_links = extra_links_for_iteration(link_type.value)
+    download_file_mock = mocker.Mock(side_effect=side_effect_func_for_download_file)
+    mocker.patch('harmony.client.Client._download_file', download_file_mock)
```
flamingbear, line 1400: So when you run this locally it generated a file in /tmp/fake.tif that file was actually the login page for EDL or if you had a .netrc it was ```{"code":"harmony.NotFoundError","description":"Error: The requested page was not found."}%```. 

This should stop accessing live sites and get the tests to pass every time.

harmony/request.py
flamingbear: Sorry, you were right. 🤦

.github/workflows/tests.yml
```diff
@@ -1,13 +1,26 @@
 name: Tests
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
```
owenlittlejohns, line 6: So this got me going down a (small) rabbit hole. It's probably worth reviewing the protections on `main` in the repository settings.

.github/workflows/tests.yml
```diff
@@ -1,13 +1,26 @@
...
+  pull_request:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 2 * * 1,3,5'
```
owenlittlejohns, line 10: I like this scheduled execution - I know we discussed it a lot a couple of weeks back, but it definitely makes me _much_ more comfortable with removing the upper limits on dependencies.

.github/workflows/tests.yml
```diff
@@ -1,13 +1,26 @@
 name: Tests
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
```
owenlittlejohns, line 8: Slightly (but not very) curious why you want the `workflow_dispatch` trigger, when it wasn't there before.

.github/workflows/tests.yml
```diff
@@ -31,3 +50,23 @@ jobs:
...
       if: ${{ github.event_name == 'pull_request' && (github.event.action == 'opened' || github.event.action == 'synchronize') }}
+
+  notify_nightly_failure:
+    needs:
+      - build
+    runs-on: ubuntu-latest
+    if: >
+      always() &&
+      needs.build.result == 'failure' &&
+      github.event_name == 'schedule'
```
owenlittlejohns, line 61: Okay, yeah. I think this makes sense to only do when running on the scheduled events.

harmony/request.py
```diff
@@ -216,7 +216,7 @@ def is_wkt_valid(wkt_string: str) -> bool:
         # Attempt to load the WKT string
         loads(wkt_string)
         return True
-    except (ShapelyError, ValueError) as e:
+    except (ShapelyError, ValueError, NotImplementedError) as e:
```
owenlittlejohns, line 219: Reading the conversation between you and @chris-durbin - my understanding here is that the `NotImplementedError` is something in newer versions of `shapely` (>=2.0.7), and wasn't being guarded against?

requirements-min.txt
owenlittlejohns, line 1: So, I think this is probably the best solution, but it feels a bit clunky to have two requirements files in the repository. Part of me was toying with whether it would be cleaner to have a rule in the `Makefile` to do some substitution equivalent to `s/\~=/==/`, but I don't think that's foolproof. So this is probably the best option.

Maybe it would help to add a comment in the `pyproject.toml`, just above `dependencies` saying someone should update this file if they also update the dependencies, too.

owenlittlejohns (approved):
This makes sense to me. The only nitpicky thing I have is possibly adding a comment in the `pyproject.toml` pointing developers at the minimum requirements file (I worry that they'll easily get out of sync).

.github/workflows/tests.yml
```diff
@@ -31,3 +50,23 @@ jobs:
         name: code-coverage-report ${{ github.event.pull_request.head.sha }} ${{ matrix.python-version }}
         path: htmlcov/*
       if: ${{ github.event_name == 'pull_request' && (github.event.action == 'opened' || github.event.action == 'synchronize') }}
+
+  notify_nightly_failure:
+    needs:
+      - build
+    runs-on: ubuntu-latest
+    if: >
+      always() &&
```
owenlittlejohns, line 59: ~~Just coming back to this:~~

~~What's the `>` mean? Should it be a `|`? (Not spotting something quickly from the docs)~~

Just found the answer - I just didn't know enough about YAML.

harmony/request.py
flamingbear: Yup. Exactly that.  Chris was just pointing out that I had my exceptions wrong and doubled up the ShapelyError.

.github/workflows/tests.yml
flamingbear: Just adding it so you can test from the interface.  No other reason.

.github/workflows/tests.yml
flamingbear: I actually think it could go even less often but this should be fine until it's broken and we can't fix it right away.

.github/workflows/tests.yml
flamingbear: Added and will update after the checks run with the new matrix on main.

requirements-min.txt
flamingbear: e2203ca

PR #130, opened 2026-06-01
fix: improve error message when submit() receives a non-JSON response

stark256-spec:
Fixes #129.

## Problem
When `harmony_client.submit()` is called but authentication is missing or broken, Harmony redirects the request to the Earthdata Login page — returning HTML instead of JSON. The `OgcBaseRequest` code path called `response.json()` without a try/except, so users saw a raw `JSONDecodeError: Expecting value: line 1 column 1` with no indication of what actually went wrong.

## Fix
Wrap `response.json()` in a `try/except requests.exceptions.JSONDecodeError` for the `OgcBaseRequest` path and raise a descriptive `Exception` that:
- Identifies the likely cause (authentication failure)
- Tells the user to check their `.netrc` file for the Earthdata Login host
- Includes the raw response status code and first 200 characters for further debugging

**Before:**
```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

**After:**
```
Exception: Harmony returned a non-JSON response. This usually indicates an
authentication failure. Check that your .netrc file contains valid
credentials for the Earthdata Login host (e.g.
urs.earthdata.nasa.gov or uat.urs.earthdata.nasa.gov).
Raw response (200): <!DOCTYPE html>...
```

PR #137, opened 2026-06-04
Fix: Improve message when submit() receives a non-JSON response

### LINKED ISSUE #129: Improve error message when `harmony_client` submission fails
I was running through the [`nisar-py` Harmony regression test](https://github.com/nasa/harmony-regression-tests/blob/main/test/nisar-py/nisar-py_Regression.ipynb), but kept failing with the error message below. 

I realized that my `.netrc` file didn't have an entry for `uat.urs.earthdata.nasa.gov`, which resolved the issue. But it would be helpful if the error message were more descriptive, letting users know what went wrong. 

Thank you! 

```
--------------------------------------------------------------------------
JSONDecodeError                           Traceback (most recent call last)
File ~/opt/miniconda3/envs/papermill-nisar-py/lib/python3.12/site-packages/requests/models.py:976, in Response.json(self, **kwargs)
    975 try:
--> 976     return complexjson.loads(self.text, **kwargs)
    977 except JSONDecodeError as e:
    978     # Catch JSON-related errors and raise as requests.JSONDecodeError
    979     # This aliases json.JSONDecodeError and simplejson.JSONDecodeError

File ~/opt/miniconda3/envs/papermill-nisar-py/lib/python3.12/json/__init__.py:346, in loads(s, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    343 if (cls is None and object_hook is None and
    344         parse_int is None and parse_float is None and
    345         parse_constant is None and object_pairs_hook is None and not kw):
--> 346     return _default_decoder.decode(s)
    347 if cls is None:

File ~/opt/miniconda3/envs/papermill-nisar-py/lib/python3.12/json/decoder.py:338, in JSONDecoder.decode(self, s, _w)
    334 """Return the Python representation of ``s`` (a ``str`` instance
    335 containing a JSON document).
    336
    337 """
--> 338 obj, end = self.raw_decode(s, idx=_w(s, 0).end())
    339 end = _w(s, end).end()

File ~/opt/miniconda3/envs/papermill-nisar-py/lib/python3.12/json/decoder.py:356, in JSONDecoder.raw_decode(self, s, idx)
    355 except StopIteration as err:
--> 356     raise JSONDecodeError("Expecting value", s, err.value) from None
    357 return obj, end

JSONDecodeError: Expecting value: line 1 column 1 (char 0)

During handling of the above exception, another exception occurred:

JSONDecodeError                           Traceback (most recent call last)
Cell In[5], line 1
----> 1 harmony_client.submit(
      2     harmony.Request(
      3         **test_case,
      4         **harmony_request_params,

File ~/opt/miniconda3/envs/papermill-nisar-py/lib/python3.12/site-packages/harmony/client.py:548, in Client.submit(self, request)
    546 if response.ok:
    547     if isinstance(request, OgcBaseRequest):
--> 548         if response.json()['status'] == 'successful':
    549             return response.json()
    550         else:

File ~/opt/miniconda3/envs/papermill-nisar-py/lib/python3.12/site-packages/requests/models.py:980, in Response.json(self, **kwargs)
    976     return complexjson.loads(self.text, **kwargs)
    977 except JSONDecodeError as e:
    978     # Catch JSON-related errors and raise as requests.JSONDecodeError
    979     # This aliases json.JSONDecodeError and simplejson.JSONDecodeError
--> 980     raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)

JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

flamingbear:
Pulling in user contribution: #130 

Fixes https://github.com/nasa/harmony-py/issues/129.

Problem

When harmony_client.submit() is called but authentication is missing or broken, Harmony redirects the request to the Earthdata Login page — returning HTML instead of JSON. The OgcBaseRequest code path called response.json() without a try/except, so users saw a raw JSONDecodeError: Expecting value: line 1 column 1 with no indication of what actually went wrong.
Fix

Wrap response.json() in a try/except requests.exceptions.JSONDecodeError  and raise a descriptive Exception that:

    Identifies the likely cause (authentication failure) 
    Tells the user to check their .netrc file for the Earthdata Login host
    Includes the raw response status code and first 200 characters for further debugging
    

Before:

```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

After:

If the non-JSON respose doesn't have the words "Earthdata Login", we just return a short message

``` 
Harmony returned a non-JSON response. [200 characters of the response]
```

If the response looks like a login page, via the fact it contains the words "Earthdata Login" on it.

``` 
Exception: Harmony returned a non-JSON response. This may indicate an
authentication failure. Check that your .netrc file contains valid
credentials for the Earthdata Login host (e.g.
urs.earthdata.nasa.gov or uat.urs.earthdata.nasa.gov).
Raw response (200): <!DOCTYPE html>...
```


tests/test_client.py
```diff
@@ -1601,6 +1601,45 @@ def test_handle_error_response_invalid_json():
...
+    assert 'Harmony returned a non-JSON' in error_msg
+    assert 'netrc' in error_msg
+    assert 'urs.earthdata.nasa.gov' in error_msg
+
+    with pytest.raises(Exception) as exc_info:
+        Client(should_validate_auth=False).submit(request)
+    error_msg = str(exc_info.value)
+    assert 'Harmony returned a non-JSON' in error_msg
+    assert 'netrc' not in error_msg
+    assert 'urs.earthdata.nasa.gov' not in error_msg
```
owenlittlejohns: Nit: This fits with the structure of other tests in this file, and how the `responses` package works, but it's a little bit non-intuitive at a first glance that the same request being submitted in the same way has a different outcome. Not sure what I'm suggesting here. Maybe separate tests, and maybe the `Request` could be a fixture shared between them?

That's possibly overkill, though. Clearly I got to the point where I understood the test, so maybe that's the best sign that no change is needed.

harmony/request.py
```diff
@@ -220,6 +220,9 @@ def is_wkt_valid(wkt_string: str) -> bool:
         # Handle WKT reading errors and invalid WKT strings
         print(f"Invalid WKT: {e}")
         return False
+    except (ShapelyError, NotImplementedError) as e:
+        print(f"Invalid WKT: {e}")
+        return False
```
owenlittlejohns: I'm not sure I'm understanding something properly here.

1) If there's a `ShapelyError`, won't it hit the previous `except` block and not get down to here.
2) Given that the content of this `except` block is identical to the previous one, couldn't you just have: `except (ShapelyError, ValueError, NotImplementedError) as e:` in one block?

harmony/client.py
```diff
@@ -545,10 +545,25 @@ def submit(self, request: BaseRequest) -> any:
...
+                            "Harmony returned a non-JSON response. This may indicate an "
+                            "authentication failure. Check that your .netrc file contains valid "
+                            "credentials for the Earthdata Login host (e.g. "
+                            "urs.earthdata.nasa.gov or uat.urs.earthdata.nasa.gov).\n"
+                            f"Raw response ({response.status_code}): {response.text[:200]}"
+                        ) from e
+                    else:
+                        raise Exception(
+                            f"Harmony returned a non-JSON response: {response.text[:200]}"
+                        ) from e
```
owenlittlejohns: This all makes sense, but it does give this block of code a fairly high cyclomatic complexity. Maybe it would be worth pulling out your new code as a function that tries to get the `response.json()` output, and handles the different exceptions. It could also be used in the `else` block below for non `OgcBaseRequest` requests.

owenlittlejohns (changes requested):
Nice work - the changes make sense and should improve the messaging back to the end-user. Nice!

I only had nits, and they aren't earth shattering. (Feel free to push back on the comment on the tests, for sure)

harmony/request.py
flamingbear: This was duplicate code to keep the tests passing that Chris pointed out in #133 and I'll pull in that correct change here.

harmony/client.py
flamingbear: I extracted it into a separate function, but I'm not 100% sure it's the right thing to do anymore after looking at how non `OgcBaseRequest` are handled.
We're failing to decode an expected json body. and we determined that it's sending HTML of the earthdata login page sometimes, and I just added the additional case for if different html came back without really knowing if it could.

tests/test_client.py
flamingbear: I moved this into two tests.

harmony/client.py
flamingbear: Ok, I talked to Chris and neither of us are very familiar but we think the OgcBaseRequest should always be an error if a non-html response is returned. But that the other could return html that is not an error, so just extracting the error handling should be good enough here and we leave the else clause as it is.

harmony/util.py
```diff
@@ -23,3 +27,35 @@ def s3_components(url: str) -> tuple:
...
+    Returns:
+        The parsed JSON body as a Python object.
+
+    :raises
+        Exception: If the response body is not valid JSON. Provides a more specific
+        message if the response appears to be an Earthdata Login authentication page.
+    """
+    try:
+        body = response.json()
+        return body
```
owenlittlejohns: Pedantic nit (ignore if you like): Couldn't you just do `return response.json()`?

owenlittlejohns (approved):
This looks good to me - thanks for addressing my comments. I had one super-duper nit, but it's not a showstopper at all.

harmony/client.py
owenlittlejohns: Okay - that makes sense to me. Thanks for extracting the helper function, even if it's just used in the one place. It does make the code a lot easier to read.

PR #138, opened 2026-06-05
Deprecate python 3.9

flamingbear:
## Jira Issue ID

None

## Description
Removes support for python 3.9 which is EOL.

## Local Test Steps

See that the actions run and the tests pass (they wont until the other PRs are merged)

#132 is needed to support python 3.14.

## PR Acceptance Checklist
* [ ] Acceptance criteria met


owenlittlejohns (approved):
These changes make sense to me, and I definitely think it is reasonable to deprecate 3.9 as it's EoL.

(Saying the obvious thing out loud) When this gets released, it should probably be a major version, just in case there are folks with notebooks on Python 3.9 out there.

PR #142, opened 2026-06-24
Improvement: Switch to git tags for version during packaging.

flamingbear:
## Jira Issue ID

HARMONY-FIX

## Description

Derive the package version from the git tag at build time usind `setuptools_scm`
so releases no longer require CI to bump and force push to `main`.

This will allow us to add branch protections to the default branch without
having to add an application to add an exception to the rules. And we can keep
the same basic workflow.

Note: `harmony/__init__.py` will no longer have a literal version string.

## Local Test Steps

This is a little tough to test, but you can do a few things.

You check out this commit and build the package with `make build`

You will get:
```
Successfully built harmony_py-1.4.1.dev1+gac794a9bf.tar.gz and harmony_py-1.4.1.dev1+gac794a9bf-py3-none-any.whl
```

so `+gac794a9bf` == built from git commit.

Or you can use a pretend version:
```
❯ SETUPTOOLS_SCM_PRETEND_VERSION=2.4.0 make build
```

and get
```
Successfully built harmony_py-2.4.0.tar.gz and harmony_py-2.4.0-py3-none-any.whl
```

you can check the `harmony.__version__` still works on either of these by installing the wheel and checking the version

```
❯ python -c "import harmony; print(harmony.__version__)"
```

You can actually fake install any version with the make command:
```
❯ SETUPTOOLS_SCM_PRETEND_VERSION=3.0.3 make install
```
...Successfully installed harmony-py-3.0.3

```
❯ python -c "import harmony; print(harmony.__version__)"
3.0.3
```

The final way you can test how CI/CD will respond is by adding a local tag in the form v.X.Y.Z on the checked out branch with everything clean, then a make build will generate a vX.Y.Z package.

## PR Acceptance Checklist
* [ ] ~Acceptance criteria met~
* [ ] ~Tests added/updated (if needed) and passing~
* [ ] ~Documentation updated (if needed)~

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 # https://pypi.org/pypi?%3Aaction=list_classifiers
 
 [build-system]
-requires = ["setuptools"]
+requires = ["setuptools>=64", "setuptools_scm>=8"]
 build-backend = "setuptools.build_meta"
 
 [project]
@@ -82,9 +82,8 @@ examples = [
     "rasterio ~= 1.3"
 ]
 
-[tool.setuptools.dynamic]
-# Will read __version__ from harmony.__init__.py
-version = {attr = "harmony.__version__"}
+[tool.setuptools_scm]
+# Version is derived from the git tag (e.g. v1.4.0 -> 1.4.0) at build time.
 
 [tool.setuptools.packages.find]
 exclude = ["contrib", "docs", "tests*"]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -2,7 +2,7 @@
 # https://pypi.org/pypi?%3Aaction=list_classifiers
 
 [build-system]
-requires = ["setuptools"]
+requires = ["setuptools>=64", "setuptools_scm>=8"]
```
flamingbear, line 5: dynamic version comes from setuptools_scm now

chris-durbin (approved):

indiejames (approved):

PR #144, opened 2026-07-04
docs: clarify download_all() is a generator, not a list

### LINKED ISSUE #140: Improve documentation for `download` and `download_all` to reflect present behavior of these methods
Hi!

I love the `harmony-py` library and share it with many people trying to access NASA data. We (NSIDC) just did a webinar short for NASA on this! There is, however, some confusing behavior in the way the download methods are working.

For example:

In the following, downloading does not start until `result` is called.
```
futures = harmony_client.download_all(job_id, directory='/tmp', overwrite=True)
filelist = [f.result() for f in futures]
```
Often, people do not include something comparable to that second line and receive no files.

In a discussion with a colleague, I learned that in v0.2.0, the download methods were turned into generator functions and don't actually return a list of futures as the documentation says. I understand that this has been this way for a while, but new users to `harmony-py` hit a stumbling block by using these methods and not calling `result`.

Thanks!

karlhillx:
## Summary

`download_all()` has been a generator since v0.2.0, but the docstrings, documentation, and example notebooks still describe it as returning a "list of Futures". This is a silent gotcha — new users call `download_all()` without iterating the result and receive no files, as reported in #140.

## Changes

- **`harmony/client.py`**: Rewrote `download_all()` docstring to explicitly state it is a **generator**, that downloads are lazy (only begin when the generator is iterated), and show the correct `[f.result() for f in ...]` usage pattern. Changed `Returns:` to `Yields:`. Also improved `download()` docstring with a usage example showing `.result()`.
- **`docs/index.rst`**: Fixed the quick-start example on the landing page to consume the generator (`filenames = [f.result() for f in harmony_client.download_all(job_id)]`) instead of calling `download_all()` without iterating.
- **`examples/basic.ipynb`**: Updated markdown to clarify `download_all()` returns a **generator** of futures, added a note about lazy iteration.
- **`examples/intro_tutorial.ipynb`**: Clarified that `download_all()` is a generator and downloads only begin on iteration.
- **`examples/job_results.ipynb`**: Fixed comment from "returns immediately with a list of concurrent.futures" to "is a generator that yields concurrent.futures.Future objects".

## The problem

From #140:

> In the following, downloading does not start until `result` is called.
> ```
> futures = harmony_client.download_all(job_id, directory="/tmp", overwrite=True)
> filelist = [f.result() for f in futures]
> ```
> Often, people do not include something comparable to that second line and receive no files.

The method signature already declares `-> Generator[Future, None, None]`, but every human-facing description said "list". This PR fixes the descriptions to match the actual behavior.

Fixes #140.

## Checklist

- [x] Documentation updated
- [x] No code changes (docs-only)
- [x] Examples use correct `[f.result() for f in ...]` pattern

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,7 +25,7 @@ Harmony-Py is a work-in-progress, is not feature complete, and should only be us
 
     >>> job_id = harmony_client.submit(request)
 
-    >>> harmony_client.download_all(job_id)
+    >>> filenames = [f.result() for f in harmony_client.download_all(job_id)]
 
 -------------------
 
--- a/examples/basic.ipynb
+++ b/examples/basic.ipynb
@@ -145,11 +145,11 @@
    "id": "threatened-complement",
    "metadata": {},
    "source": [
-    "There are a number of options available for downloading results. We'll start with the 'download_all()' method which uses a multithreaded downloader and quickly returns with a \"future\" (specifically a python conccurrent.futures future).\n",
+    "There are a number of options available for downloading results. We'll start with the 'download_all()' method which uses a multithreaded downloader and returns a **generator** of futures (specifically python concurrent.futures objects).\n",
     "\n",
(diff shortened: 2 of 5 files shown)


karlhillx:
Closing in favor of #143, which was opened first and has Copilot review feedback. The docstring, docs, and notebook fixes from this PR have been folded into #143.

